PR #10, opened 2021-04-12
Wisp Suggestion: End Day

McTsts:
I propose we change Wisp to be an Immediate Day effect instead of an Immediate Night effect. 
a) In OS6C it was kind of weird for Wisp to have an immediate ability that they'd have to use as late as possible into the night
b) It's a buff but I don't really think it'd be an issue. If this is used in a full game it's more likely there's 1 Pyromancer and 1 Wisp, so communication between them is probably pretty difficult anyway

Alternatives: Make it an end night ability


venomousbirds:
Okay that's interesting. I don't think it'll do harm, but I do think we should test it in a few games before changing it because I think it's quite a large change

McTsts:
yeah probably should be tested, but the immediate night thing just seems weird to me. The more obvious solution would be end night ofc, but I definitely thing immediate day could be more interesting.

venomousbirds:
We could try out both in an upcoming game and see what works best

McTsts:
yeah

Mjnstag:
Sound good, but yea, probably should be tested. Might be interesting in the case that they do find each other since then they basically have a barber ability which they could use to influence poll results (e.a. killing person who would otherwise be elected mayor and then sniping mayor themselves?)

McTsts:
yeah immediate day kills can be pretty useful, though sniping is unlikely for a Solo team

McTsts:
this is a draft now until we've tried it

McTsts:
This was changed to end day as requested by luffy(?)

venomousbirds (approved):

PR #17, opened 2021-04-13
Coroner Buff

McTsts:
Coroner was removed because it was too weak. Making it a passive power would be a buff and would make it more useful.

### DIFF
--- a/limited/coroner
+++ b/limited/coroner
@@ -2,5 +2,5 @@
 __Basics__
 Upon every death, the Coroner is informed how that participant died. 
 __Details__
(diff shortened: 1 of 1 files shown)


venomousbirds:
You mean it gets to know the cause of every single death, kind of like a reporter but for causes?

venomousbirds (commented):
I don't know if it worked but I changed something

McTsts:
> You mean it gets to know the cause of every single death, kind of like a reporter but for causes?

yeah

> I don't know if it worked but I changed something

It worked. It's definitely better, thanks!

McTsts:
I added a clarification: Coroner can't be affected by disguises

venomousbirds (approved):

PR #45, opened 2021-04-13
Trickster Role Loss

### LINKED ISSUE #25: Trickster: missing sentence in bot
"If the Trickster loses their role, their vote will count as 1." not in bot description

McTsts:
Fixes #25 
I changed the sentence slightly, because usually it's better to avoid these statments with absolute numbers. (e.g. 'vote will count as 1' makes it unclear how that works with Mayor or Grandma, etc)


Mjnstag:
looks good

PR #48, opened 2021-04-13
RH version from role book

### LINKED ISSUE #39: Riding hood: different description
The descriptions from the bot and rolebook are completely different with the rolebook having a more extensive explantaion about the role?

McTsts:
Using the longer description from the role book instead.
Fixes #39


Mjnstag:
👍

PR #56, opened 2021-04-14
Devil cant be wager killed

### LINKED ISSUE #36: Devil: wager kill the devil
"Using the wager to attack the Devil always fails."

Maybe we should add something to clarify that when a wager is used to kill the devil that the devil wont have an increase in souls like when attacking another player since the soul is immediately used to protect the devil.

McTsts:
Fixes #36

### DIFF
--- a/other/hell/devil
+++ b/other/hell/devil
@@ -9,7 +9,7 @@ The Devil can't target soulless players. The wager is sent at the end of the day
 The Devil can anonymously communicate with the target & will be notified of the result of the wager - they will learn who was attacked/which person was inspected & what role the target saw. 
 
 If the player chooses to see another player’s role (immediate), it will be affected by weak disguises and obstructions. These effects are seen from the player’s perspective, not the Devil’s.
-If the player chooses to attack another player (soul sold: immediately; kill executed: end-night), then they will become soulless. The attacker is the Devil. Using the wager to attack the Devil always fails.
+If the player chooses to attack another player (soul sold: immediately; kill executed: end-night), then they will become soulless. The attacker is the Devil. Using the wager to attack the Devil will fail end-night, but the soul is still sold.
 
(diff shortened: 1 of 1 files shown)


Mjnstag:
maybe better to also mention that the devil will (not) gain a soul since that might be vague from that change?

Mjnstag (changes requested):
maybe better to also mention that the devil will (not) gain a soul since that might be vague from that change?

venomousbirds (changes requested):
It currently seems like the attack on the devil fails but the devil still gains a soul, I don't know how to word it better but right now it's worded even weirder

Mjnstag:
maybe something like:
"Using the wager to attack the Devil will fail end-night, but the Devil will still gain a soul."

or:
"Using the wager to attack the Devil will fail end-night, but the Devil will still gain the sold soul."

or:
"Using the wager to attack the Devil will fail end-night, but the sold soul will be used to protect the Devil."

venomousbirds:
Last option is the only option that actually clarifies that the sold soul is not added to the current Devil souls

McTsts:
The sold soul **is** added to the Devils total.

venomousbirds:
It's added and removed. Right now it sounds like it's not removed

McTsts:
It is **not** removed.

McTsts:
Once the soul is sold, the wager's recipient no longer has anything to do with anything that happens. The Devil executes the devil's wager attack. The Devil will not attack themself. The Devil gains the soul from the wager and just doesn't do the attack.

venomousbirds:
Makes sense in that case, I just remember some discussion that the soul gaining specifically shouldn't be added for some reason

McTsts:
@venomousbirds @Mjnstag I've changed it and it should be clearer now.

venomousbirds (approved):

Mjnstag (approved):

PR #58, opened 2021-04-14
Pack Tanner as Limited role

McTsts:
This isn't anything super fancy so I don't think accepting this as a limited role would be a problem

### DIFF
--- a/limited/pack tanner
+++ b/limited/pack tanner
@@ -2,5 +2,7 @@
 __Basics__
 Each day, the Pack Tanner may disguise up to one players as a role of the Pack Tanner’s choosing.
 __Details__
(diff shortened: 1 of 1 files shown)


venomousbirds (approved):
Might even have potential as a slight buff to the wolfpack in big games?

McTsts:
well yeah

venomousbirds:
I meant as in, while it's a limited role, it has potential to become an actual recognized role

McTsts:
oh yeah definitely. Just I think WIP roles usually are Limited ones first.

venomousbirds:
Yeah that's true and it should be that way, I'm just acknowledging that Pack Tanner sounds like a great addition

McTsts:
Yeah, we could try using it next game. I think using a few Limited roles in normal games is probably fine.

Dennis-0 (approved):

PR #59, opened 2021-04-14
Lab Assistant as Limited Role

McTsts:
Evil Scientist isn't a very commonly used Solo role, but I still think it makes sense to add the weaker form as a limited role


Mjnstag:
yea, especially since we have a weaker role for devil

venomousbirds (approved):

Dennis-0 (approved):

PR #60, opened 2021-04-14
Exorcist as limited role

McTsts:
Exorcist should be added as a limited role, though the description still needs to be improved a bit I think.


venomousbirds (changes requested):
Not clear what exactly is seen as a negative effect

Mjnstag:
maybe it should affect negative wolf effect like being threatened? but that might unbalance things

McTsts:
you mean being sabotaged?

Mjnstag:
nvm, confused stalker with wolfpack

McTsts:
> Not clear what exactly is seen as a negative effect

I've added a list of negative effects.

McTsts:
This one is probably not gonna ever be a full role, but I think it can be useful for solo focused off-season and mini games

venomousbirds:
Still mentions frozen, is that neccesary? (Ice King removed, i dont even know what we're doing with IQ)

McTsts:
> Still mentions frozen, is that neccesary? (Ice King removed, i dont even know what we're doing with IQ)

Both are limited roles now. I've moved the frozen effect down to the lower section (the one for effects of limited roles)

venomousbirds (approved):

Mjnstag:
> he Exorcist can not cure players if they have changed role as the result of an effect, such as the Undead, but can still remove the effects the Undead has. 

would that mean that, in the case above, that an undead wolf gets its private vote back and can vote on wolfpack polls but still win with vampire?

Also, would curing an enchanted player mean that they also get removed from the enchanted channel? they get added to it when they are enchanted, doesnt mention if they should be removed if they are not enchanted anymore.

McTsts:
> > he Exorcist can not cure players if they have changed role as the result of an effect, such as the Undead, but can still remove the effects the Undead has.
> 
> would that mean that, in the case above, that an undead wolf gets its private vote back and can vote on wolfpack polls but still win with vampire?
> 
> Also, would curing an enchanted player mean that they also get removed from the enchanted channel? they get added to it when they are enchanted, doesnt mention if they should be removed if they are not enchanted anymore.

That sentence is very confusing, what I meant is that the Exorcist can remove powdered/enchanted/demonized from the Undead but they remain Undead (and still don't have a vote)

Exorcising an enchanted player should remove them from the enchanted channel.

Mjnstag:
> remove (...) demonized from the Undead

can they be demonized again in that case?

McTsts:
> > remove (...) demonized from the Undead
> 
> can they be demonized again in that case?

Yes, but only if they are no longer part of the underworld team.

Dennis-0 (approved):

PR #61, opened 2021-04-14
Disguised Fox default disguise consistency

McTsts:
I think the default disguise should probably be Citizen instead of Wolf


Mjnstag:
It would make it more difficult for wolfpack to find df with psychic wolf/warlock but it probably would be better for df if they forget and get checked by town. So idk

McTsts:
I mean this **should** not actully be a relevant change. Also, I don't think any PW would try to check for 'Wolf' to find a DF that forgot to disguise. Warlock is very slightly nerfed in a super specific situation (DF misses one disguise, Warlock finds them, DF joins pack to let them know who Warlock is) - but I dont think this actually really matters

venomousbirds:
I feel like citizen might nerf it?

McTsts:
How is Citizen a worse disguise than Wolf though

venomousbirds:
People don't quickly check for wolf and citizen is often mistrusted. I do understand the buff but there's already a wolf role that's disguised as citizen

McTsts:
@venomousbirds This is just the default disguise if they don't choose one. I don't understand what your point is to be honest. Yes, a CS probably isn't looking for Wolf but also probably not for Citizen so it doesn't make a difference there. Huntsman, AT, Oracle and FT all get a better result with Citizen then they get with Wolf.

venomousbirds:
I just don't think the change is neccesary, it's fine the way it is

McTsts:
@venomousbirds I really don't understand the reasoning for this. Why is a Wolf disguise better than a Citizen one in _any way_? If the DF for some reason wants to be Wolf they can still choose that disguise.

venomousbirds:
I just think it doesn't make sense for a wolf role to be disguised as a citizen if it isn't specifically their role to be citizen (sneaking wolf). Keeping it Wolf encourages the Disguised Fox to change their disguise too

McTsts:
My proposed change is consistent with how Devil (and Dog) work. Devil pretty much has the same situation where they choose a disguise and if they don't they still get a _usable_ one. There's no reason to be punishing for no reason.

Mjnstag (approved):

McTsts:
@venomousbirds !

venomousbirds (changes requested):
Please keep it wolf
thight

McTsts:
... but why

venomousbirds (changes requested):
I won't approve of this, I've explained why

Shepherdwhen:
Idk why it wold default to wolf, since wolves already know DF's disguise and it's more helpful to default to cit if they don't disguise so they don't get checked, unless you want to punish ans inactive DF, but just find a substitute at that point.

venomousbirds:
They can change it away from wolf on d0 anyways, so it's really not worth changing it to citizen

McTsts:
> I won't approve of this, I've explained why

And I've explained why your argument doesn't make sense as we handle the same situation differently elsewhere.

If you look at other disguises, for all disguises where the player can choose either "Citizen" disguise is applied if none is chosen (as Citizen is considered the default role - see also Dog) or alternatively no disguise is applied at all.

Devil/Archangel/Shepherd => Citizen
Psychopath/Tanner/Pack Tanner => None

> I just think it doesn't make sense for a wolf role to be disguised as a citizen if it isn't specifically their role to be citizen (sneaking wolf).
I just think it doesn't make sense for a Disguised Fox to be disguised as wolf if it isn't specifically their role to be wolf - If we did want to punishing the DF for no reason, no disguise would probably make more sense than "wolf".
Additionally, Citizen really always make sense since it is the default role.

venomousbirds:
no disguise would work too, in that case

Shepherdwhen:
Doesn't disguised fox lose it's disguise at the end of the night? or does it stay on it until it changes?
If the former then Ts's decision is the best one
If the latter then I agree with Vera's decision

I have been discussing this assuming the former

venomousbirds:
> The Disguised Fox's disguise wears off at the end of each night.

Shepherdwhen:
> > The Disguised Fox's disguise wears off at the end of each night.


Well then Ts' idea makes most sense as you're argument only works if the disguise doesn't change from the last day if they forget on a day in the middle of the game, Ts' point means that if the DF forgets D4 it wont be disguised as wolf that night and instead will be disguised as cit

Shepherdwhen:
This was never agreed upon, I really think a natural wolf disguise is less useful than no disguise, it should be a natural citizen disguise

venomousbirds:
Still think wolf disguise makes the most sense, because DF is wolf team

McTsts:
This change is mainly about consistency in the role book and has no real relevance to how the game is played.

Shepherdwhen:
> This change is mainly about consistency in the role book and has no real relevance to how the game is played.

It's the only role with a non cit natural disguise I do believe

Shepherdwhen:
Natural being what it defaults to when no disguise is submitted

McTsts:
Yeah

Dennis-0 (approved):
I think citizen default disguise is better. Also, if a disguised fox would disguise as a wolf by default, it would basically tell the wolves that the disguised fox is inactive, which could help the wolves to know who the disguised fox is.

PR #63, opened 2021-04-14
Clairvoyant Fox draft

McTsts:
draft for a role Ive wanted to make for a while.


venomousbirds:
What ability do you think it should have?

McTsts:
It's definitely some way of investigating. I think what we originally discussed was that it checks a players role and the pack is also informed of the results.

Mjnstag:
sounds interesting. could make for an interesting dynamic between pack and fox. Maybe add a line about not being able to use it on their selves to let the pack know who they are, that would defeat the purpose of not being in the wolfpack imo

McTsts:
oh yeah didn't notice that

venomousbirds:
Added a description that might work

McTsts:
> sounds interesting. could make for an interesting dynamic between pack and fox. Maybe add a line about not being able to use it on their selves to let the pack know who they are, that would defeat the purpose of not being in the wolfpack imo

ill add this as well, but other than that yours seems good 👍

venomousbirds (approved):

Dennis-0 (approved):

PR #72, opened 2021-04-14
Mad Scientist Solo seen as different teams now

venomousbirds:
Added Solo teams as different teams but not sure if it looks good


McTsts (approved):
That works I think

PR #76, opened 2021-04-15
Update mayor

Mjnstag:
Day -> night when they choose if there will be 1 or 2 lynches


McTsts (changes requested):
should probably be clarified that they decide for the lynch on the next day & that this is announced. at that point probably also split it into several sentences

McTsts:
this should fix it

McTsts (approved):

PR #85, opened 2021-04-15
Solo category change (<Team> <Category> => Solo <Category> {- <Team> Team})

McTsts:
Some other roles still had their team listed as "solo" when they actually have their own team names.

### DIFF
--- a/limited/archangel
+++ b/limited/archangel
@@ -1,4 +1,4 @@
-**Archangel** | Heaven Killing
+**Archangel** | Solo Killing - Heaven Team
 __Basics__
 The Archangel is an alternative version of the devil.
 __Details__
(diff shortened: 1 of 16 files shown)


McTsts:
alternatively we could add the solo back to the other roles but I definitely prefer this

Mjnstag:
Maybe it would be good to mention solo somewhere since if someone new searches for the role they might think that it isnt solo, which is a problem in closed list where there is only solo: 2

venomousbirds:
We could list it as "Solo <category> - <Solo group name> Team"
for example Devil would be "Solo Killing - Hell Team"

McTsts:
that seems like a good idea

McTsts:
Changed everything to that format.

venomousbirds (approved):

Dennis-0 (approved):

PR #88, opened 2021-04-15
Child/RK Edge Case moved to RK

McTsts:
Moved the Child/Royal Knight edge case from the Child description to the Royal Knight description. It was originally in the Child description as the Child was a limited role and hence the edge case couldn't be in the RK description.


venomousbirds (approved):
Makes more sense to be in the RK description

PR #92, opened 2021-04-15
Update town crier

### LINKED ISSUE #91: Town Crier doesnt know player
Town Crier knows the identity and role of another player (and not just the role) - this should be clarified

McTsts:
Fixes #91 
I also changed "random" to "arbitrary" because if GMs can decide it can make the role list less volatile (doesn't mean you cant randomize anymore either)


venomousbirds (approved):
Had to look up what arbitrary meant but it's defintily more accurate this way

PR #98, opened 2021-04-15
Create Caroler

Shepherdwhen:


McTsts (changes requested):
Seems like the role is incomplete? Also, please don't put add empty lines.

venomousbirds:
Too many requirements, the long sentence is incomprehensible and hard to read for me. I don't like this idea because it doesn't actually investigate, it just gives a bunch of random names. Seems too easy

McTsts:
Seems unfinished. Re-open this if you feel like finishing it.

PR #102, opened 2021-04-16
Cupid change

### LINKED ISSUE #122: Cupid "role loss"
is missing (by that I mean its missing from the list)

McTsts:
This clears up some Cupid related edge cases 
Fixes #122


venomousbirds:
> If the cupid has a soul they lose it the moment they become a Demonic Cupid.

Does the Devil gain that soul?

McTsts:
> > If the cupid has a soul they lose it the moment they become a Demonic Cupid.
> 
> Does the Devil gain that soul?

The soul is supposed to be lost. (Right now Cupid would turn into (soulless) Demon so this is consistent with how it is currently) I'll clarify that.

unaligned/align/cupid
```diff
@@ -4,4 +4,6 @@ The Cupid chooses a player to be their lover, and forms an inseparable team with
...
 The Cupid and their lover form an inseparable team; they will die if the other player dies. Choosing a lover is an immediate ability.
-By default the Cupid wins with the townsfolk team. If the lover is on a different team, they will only win when both win conditions are fulfilled. For example, if the couple is Cupid and a werewolf, everyone else must die. Another example is when the couple is Undead and a Vampire, then the normal Vampire win condition must be reached. Or, if they’re both townsfolk, they belong to the townsfolk team and can win together with them. The Cupid’s love overrides any other mechanic that may cause a player to change teams for any reason; the Cupid is always on the same team as their lover. After choosing a lover, unless they are also unaligned, the Cupid is no longer unaligned.
+By default the Cupid wins with the townsfolk team. If the lover is on a different team, they will only win when both win conditions are fulfilled. For example, if the couple is Cupid and a werewolf, everyone else must die. If the lover is townsfolk, they belong to the townsfolk team and can win together with them.
```
cryptidjayden, line 7: should the cupid be default town-aligned? seems to punish wolves.

also, even if so, the "default" being town now only applies to town and wolves, so is no longer really default. would suggest phrasing change.

McTsts: The default still applies to other solo teams (Hell has an exception via Demonic Cupid - Undead Cupid is different from it as Undead Cupid's lover can be on any team)

Shepherdwhen: When a Cupid targets a wolf it basically turns it into a white werewolf without the extra kill and with a liability that it can't tell wolves about so it's really a nerfed www

McTsts: I'm not sure what alternatives there are? I guess just that the Cupid dies if they select a wolf?

Shepherdwhen: I think Cupid could possibly win with the alignment of whoever it targets, currently it's pretty much a loss for the Cupid if it targets a wolf or a solo since it has to outlive town, wolves and solo and as soon as wolves or solo catch on that Cupid targeted one of them then the they'll just kill Cupid, I don't know why Cupid as an Unaligned align has a base alignment if it's forced to target Night 1

McTsts: It's because Cupid originally is a Townsfolk role and it makes sense to be like this as we already have LA. Since it has a base alignment it's way less volatile.

McTsts:
I added lupine cupid as a way to make Cupid x Wolf more reasonable, but while keeping the way Cupid is supposed to work

Shepherdwhen:
> I added lupine cupid as a way to make Cupid x Wolf more reasonable, but while keeping the way Cupid is supposed to work

Kinda feels really strong now

Basically is www who wolves don't know about and they cannot kill but it can kill them

McTsts:
> > I added lupine cupid as a way to make Cupid x Wolf more reasonable, but while keeping the way Cupid is supposed to work
> 
> Kinda feels really strong now
> 
> Basically is www who wolves don't know about and they cannot kill but it can kill them

hm true. Originally I just wanted to do the invulnerability to pack but I wasn't sure if that was good enough.

Shepherdwhen:
> > > I added lupine cupid as a way to make Cupid x Wolf more reasonable, but while keeping the way Cupid is supposed to work
> > 
> > 
> > Kinda feels really strong now
> > Basically is www who wolves don't know about and they cannot kill but it can kill them
> 
> hm true. Originally I just wanted to do the invulnerability to pack but I wasn't sure if that was good enough.

The issue is if the cupid targets anything that is in or can join the pack basically the pack don't know cupid's identity or that it is lupine but the wolf can just tell the cupid all the pack members and they all get killed

McTsts:
> > > > I added lupine cupid as a way to make Cupid x Wolf more reasonable, but while keeping the way Cupid is supposed to work
> > > 
> > > 
> > > Kinda feels really strong now
> > > Basically is www who wolves don't know about and they cannot kill but it can kill them
> > 
> > 
> > hm true. Originally I just wanted to do the invulnerability to pack but I wasn't sure if that was good enough.
> 
> The issue is if the cupid targets anything that is in or can join the pack basically the pack don't know cupid's identity or that it is lupine but the wolf can just tell the cupid all the pack members and they all get killed

yeah making it a www is probably too much.

McTsts:
Removed the WWW part

McTsts:
I cleaned up Lupine Cupid and added role category stuff.

McTsts:
Summary of what this does:

- Demonic/Saintly Cupid: This a clarification for how Hell x Cupid works, it's rather unclear, but this should be how it already works currently. Additionally, hell team would gain the Cupid's soul, this could both be seen as a buff/nerf because they get the soul without having to use a wager (buff), but on the other hand the Cupid could've used the wager for a 2nd kill (nerf)
- Undead/Zombified Cupid: No change to how it works currently, literally just for clarification.
- Lupine Cupid: Buff for a WolfxCupid couple: Wolves arent able to kill the Lupine Cupid
- Lover: Just a role description for the lover, no change to how anything works

venomousbirds (approved):

Shepherdwhen:
Do we make one for flock if there is one for plague and archa, also is new nightmare team 100% trust?

McTsts:
> can the demonic cupid use the pool of souls? and im guessing it gets added to the hell channel?

The Demonic Cupid can't access the pool of souls.

Shepherdwhen:
> > Do we make one for flock if there is one for plague and archa, also is new nightmare team 100% trust?
> 
> Undead Cupid exists to replace the weird behavior of a Cupid turning Undead but still keeping the properties of their previous role. The actual behavior is the same, just cleaned up.
> Demonic Cupid exists because Hell is a 100% trust team.
> Saintly/Zombified exist because I already made the normal versions, so I also just made those ones.
> Flock is currently limited, and afaik doesn't have a reason to have a special cupid anyway.

I just wondered since it was 100% trust but I guess it would have a wincon chang maybe, also yeah I forgot plauge and archa are just variants not limited roles

McTsts:
Maybe we'll wait with this one, Cupid stuff is pretty complicated.

McTsts:
Updated summary, explains each Cupid.

Cupid: updated to handle UA or no lover, currently _should_ work like this but is beyond unclear
Demonic Cupid: The current situation is incredibly unclear. There are two conflicting statements, one resulting in the Cupid being unable to win when selecting a Hell team member, the other suggesting they become a part of the Hell team. This solution goes with the latter and additionally makes the Cupid soulless (not really a buff/nerf, could be seen as both in some way)
Lupine Cupid: This one is just a slight buff, and not one to resolve an issue
Undead Cupid: This one doesn't introduce any new behavior at all, just clears up the Cupid=>Undead role change.
Lover: No new info, but why not add it as a role info thing for clarification purposes
Cupid Wolf: This one doesn't introduce any new behavior at all, just clears up the Cupid=>Wolf role change.
Saintly & Zombified Cupid are just limited variants

TL;DR
Lupine Cupid: minor buff
Demonic/Saintly Cupid: clarification, minor buff/nerf
Undead/Zombified Cupid, Cupid, Lover, Cupid Wolf: clarification

MrTurtleJr:
Currently the Lupine Cupid has Townsfolk as their base alignment. Is this purposeful?

Shepherdwhen:
> Currently the Lupine Cupid has Townsfolk as their base alignment. Is this purposeful?

Yes lupine cupid + wolf works like current cupid + wolf except lupine cupid is immune to wolf kills
It is supposed to form a solo couple team

McTsts:
Yeah, that's just a minor buff to make it a bit easier.

McTsts:
Cupid Wolf is the wolf-aligned cupid variant.

PR #104, opened 2021-04-16
Update Clairvoyant Fox Typo(s)

Mjnstag:
i could only see 1 type so far


McTsts:
cant find another one either, maybe it only was 1 after all?

PR #114, opened 2021-04-17
Create ferocious fox

McTsts:
im not sure about this one yet but I had the idea so I wanted to write it down, might delete it if I dont like Clairvoyant Fox


McTsts:
I'm not really super convinced by this role so I'll close it

PR #126, opened 2021-04-17
Create blessed wolf

McTsts:
this is a role from the old role book, with the modification that the attacker learns why they escaped. (Sacred Wolf was nerfed and became Scared Wolf) Haven't thought about it too much but I'm not sure why it got removed (yes, it's stronger than a Scared Wolf, but most of the wolves usually die during the day anyway)


cryptidjayden:
sounds good

venomousbirds (approved):
it's probably a good role to put in a smaller game, to give the wolf team a little more of an advantage. Give a good variety in wolfpack roles too

McTsts:
I'm honestly not sure if this one is actually much stronger than Scared Wolf, since I added the Runner-like thing. (the original Sacred Wolf was just that it escapes any attacks), but I think the change makes it more balanced

McTsts:
If anyone has any opinion on if just escaping (without being revealed) or my version is better let me know :)

Shepherdwhen:
For a prospective name change what about "alert wolf", because sounds too much like scared wolf

Shepherdwhen:
also what about not the role that attacked it but the alignment, so it distinguishes between solo and town or would that be too weak, I think it's a little bit better than scared but not too strong that wolves use it to work out what TK is

McTsts:
Or more fittingly: "Immune Wolf" or something similiar to that

McTsts:
> also what about not the role that attacked it but the alignment, so it distinguishes between solo and town or would that be too weak, I think it's a little bit better than scared but not too strong that wolves use it to work out what TK is

Is that a reply to something? The idea seems decent, but I don't really think it fits the role.

Shepherdwhen:
Immune would suggest it is immune to all kills not just the one, alert implies that it escapes and then sees who/what/how it was attacked


> > also what about not the role that attacked it but the alignment, so it distinguishes between solo and town or would that be too weak, I think it's a little bit better than scared but not too strong that wolves use it to work out what TK is
> 
> Is that a reply to something? The idea seems decent, but I don't really think it fits the role.

Just a thought I had because the origional said it was removed, I thought it was removed because it was abused to find TK

McTsts:
I think you misread the role 
hepi

Shepherdwhen:
Oh, yep I'm dumb I thought It was told what attacked it and escaped 1 attack but it's escapes all attacks but attacker knows they are sacred wolf

McTsts:
Renamed to avoid confusion

McTsts:
This role wins the award for most disliked role suggestion 👍

McTsts:
I also voted against this role lmao

PR #128, opened 2021-04-17
Hellhound Betrayal

McTsts:
This change would make the role a bit more useful than it currently is, without really changing it too much.


venomousbirds:
disbandment sounds good to me

Shepherdwhen:
so if a hellhound becomes a vampire basically, or if an infected wolf with hellhound targets a demon/devil I guess, but I like the idea that it's a complete trust team

McTsts:
I think with normal roles the only possibilities are Undead and Cupid.

Shepherdwhen:
Oh yeah cupid turns them into solo

PR #130, opened 2021-04-17
Add Clairvoyant Fox & Fox Info Section

McTsts:
if we have several foxes it's probably worth an info section, we don't have several yet so this can remain open until we do


McTsts:
this section could also be something like "Lone Werewolves & Foxes" actually to explain the two types of special wolves that we have several of

venomousbirds:
i like the concept of foxes, and because they're quite different to pack wolves, an info section on them would be useful. we don't a lot of foxes yet however

McTsts:
> i like the concept of foxes, and because they're quite different to pack wolves, an info section on them would be useful. we don't a lot of foxes yet however

Yep, that's why this one's just a draft. If we add CF I'll make it an active PR

Shepherdwhen:
How come this PR was never closed?

venomousbirds:
It's still a draft

voidmist:
> > if we have several foxes it's probably worth an info section, we don't have several yet so this can remain open until we do
> 
> we don't have a lot of foxes yet

Soo does that mean you want more foxes? 
We have 3 foxes now though: clairvoyant, disguised and ferocious.

Shepherdwhen:
> > > if we have several foxes it's probably worth an info section, we don't have several yet so this can remain open until we do
> > 
> > 
> > we don't have a lot of foxes yet
> 
> Soo does that mean you want more foxes XD
> We have 3 foxes now though: clairvoyant, disguised and ferocious.

FF isn't a role yet just a concept for on

McTsts:
This PR makes CF a proper role and adds the fox info section. Without CF added the section doesn't make much sense and we haven't decided to add CF yet.

PR #136, opened 2021-04-17
Delete survivor

McTsts:
The role was removed and hasn't been used since. There's not really any point in keeping it.


venomousbirds (approved):
if it still stays on miniwolves, it still get used there sometimes

McTsts:
role still exists, even if it's not in the role book. Mini Wolves has a bunch of those.

PR #142, opened 2021-04-18
Update aura teller

### LINKED ISSUE #140: AT Clarification
Clarify that only roles _classified_ as Killing roles are threats.

venomousbirds:
fixes #140


McTsts:
added an extra sentence to make sure it's clear that there's some exceptions

Issue #144, opened 2021-04-19
Macho Clarification


McTsts:
clarify that it is announced when a Macho was attacked.

McTsts:
Already was fixed in that PR

PR #145, opened 2021-04-19
Update macho

venomousbirds:
not sure how to word it


venomousbirds:
should fix #144

McTsts (changes requested):
sounds pretty confusing

McTsts:
I don't think the "won't survive" thing needs clarification probably?

McTsts (approved):

PR #158, opened 2021-04-20
Update mayor

### LINKED ISSUE #107: Mayor >15
is this power actually useful?

McTsts:
Fixes #107

This sounds pretty terrible, can somebody try to improve the sentences?


Shepherdwhen:
Both mayor's or just the <15 mayor?

Shepherdwhen:
Oh wait it's just one role :s

McTsts:
Yeah, there's only one role description.

Mjnstag:
would there be any reason for the mayor to choose a different election order?

McTsts:
> would there be any reason for the mayor to choose a different election order?

they could choose Reporter in case they die

Mjnstag:
ah thats fair yea. Btw, if we accept this change, we will need to change a bunch of other roles that have the 2 mayor votes in their description like trickster/grandma

McTsts:
> ah thats fair yea. Btw, if we accept this change, we will need to change a bunch of other roles that have the 2 mayor votes mentioned in their description like trickster/grandma

This only removes the 2 lynch power and not the 2 votes power.

Shepherdwhen:
Is this basically a town flute player?

McTsts:
> Is this basically a town flute player?

Is this for Innkeeper #166

venomousbirds:
> If, by the start of the day, the chosen election isn't possible, because somebody is already elected as that role, elections will happen in the normal order.

what does this mean? it sounds a bit unclear

McTsts:
> > If, by the start of the day, the chosen election isn't possible, because somebody is already elected as that role, elections will happen in the normal order.
> 
> what does this mean? it sounds a bit unclear

Assume a Mayor and Guardian alive
Mayor chooses Guardian for the next election
Option 1) Guardian dies during the night => Guardian Election (as chosen)
Option 2) Guardian lives => Reporter Election (according to normal order)

McTsts:
Not sure how much I actually like this power. But I do think the double vote always being there makes sense.

kruthers (changes requested):
I prefer the double votes as it gives more of a balancing act and the only real reason why you may change the order is if you are like a wolf and want to push the mayor vote back by a phase

McTsts:
> I prefer the double votes as it gives more of a balancing act and the only real reason why you may change the order is if you are like a wolf and want to push the mayor vote back by a phase

With double votes, do you mean double lynch?

PR #165, opened 2021-04-21
Create assistant

McTsts:


McTsts:
I like this idea but I think there are some things that should be discussed:

- Can it redirect attacks (since attacks are a wolf power)?
- Why can it not redirect onto pack wolves? 
- If we keep that: We could consider one of the following two options:
1) It cannot redirect onto the source (so e.g. a wolfpack action cant be redirected onto wolfpack members and a CF check cant be redirected onto CF) 
2) It cannot redirect onto any wolves

venomousbirds:
sounds a little weird to limit it to not being able to redirect to wolves. what happens if the coward selects a wolf as a target? the whole point of their role is gone.

McTsts:
> sounds a little weird to limit it to not being able to redirect to wolves. what happens if the coward selects a wolf as a target? the whole point of their role is gone.

Yeah, I'm not sure as this isn't my idea, but I guess it's because it wouldn't be helpful anyway on things like invest checks and for attacks it'd be too strong?

McTsts:
Actually, I think attacks shouldn't be redirected, just so this can be a weaker role. If it can redirect attacks it'd join our group of roles that the wolves aren't safe to attack. And we already have quite a few of those. Would be better in the Citizen/Journalist group.

venomousbirds:
if stuff like invests or attacks get redirected on anything but wolves, it'd be too big of a chance to target town, which kinda makes coward seem useless

venomousbirds:
how do you propose coward could be improved to not make wolves avoid it?

McTsts:
> if stuff like invests or attacks get redirected on anything but wolves, it'd be too big of a chance to target town, which kinda makes coward seem useless

if the Coward is seen as something like FT by the wolves this would be good for town

McTsts:
> how do you propose coward could be improved to not make wolves avoid it?

Well, my latest change makes it so attacks aren't redirected. I think it makes more sense for the kind of role I want this to be. It's safe for wolves to attack, it just messes up their other powers a bit.

Shepherdwhen:
What powers would it affect PW, LW, CF, Tanner, and not even really usefully if it's not onto the pack

Also Infected wolf I would rather target coward than FT if I was town

Sab wolf is a good one if you are trying to determine leaks if you try elect the coward as mayor and make plans with the group at night

McTsts:
> What powers would it affect PW, LW, CF, Tanner, and not even really usefully if it's not onto the pack
> 
> Also Infected wolf I would rather target coward than FT if I was town
> 
> Sab wolf is a good one if you are trying to determine leaks if you try elect the coward as mayor and make plans with the group at night

The invest ones can definitely be useful - can be used to pretend to be a much stronger role and then the wolves kill you instead of that role.

Shepherdwhen:
> > What powers would it affect PW, LW, CF, Tanner, and not even really usefully if it's not onto the pack
> > Also Infected wolf I would rather target coward than FT if I was town
> > Sab wolf is a good one if you are trying to determine leaks if you try elect the coward as mayor and make plans with the group at night
> 
> The invest ones can definitely be useful - can be used to pretend to be a much stronger role and then the wolves kill you instead of that role. (and I don't see how invest powers would only be useful if redirected onto pack? in that case its just obvious it was redirected)

Only wolf investigatives to be able to show it to wolves reliably are PW and CF and PW isn't guaranteed to check you as it

McTsts:
> > > What powers would it affect PW, LW, CF, Tanner, and not even really usefully if it's not onto the pack
> > > Also Infected wolf I would rather target coward than FT if I was town
> > > Sab wolf is a good one if you are trying to determine leaks if you try elect the coward as mayor and make plans with the group at night
> > 
> > 
> > The invest ones can definitely be useful - can be used to pretend to be a much stronger role and then the wolves kill you instead of that role. (and I don't see how invest powers would only be useful if redirected onto pack? in that case its just obvious it was redirected)
> 
> Only wolf investigatives to be able to show it to wolves reliably are PW and CF and PW isn't guaranteed to check you as it

CF, LWa and W all directly see you as a role, but I guess for all of those it could actually be useful

McTsts:
One change that should be considered is that it can just redirect onto anyone.

McTsts:
> One change that should be considered is that it can just redirect onto anyone.

Did that

Mjnstag:
is the coward informed when they redirect something? it doesnt mention it in the description

venomousbirds:
Should probably be informed. Should the target be informed?

McTsts:
> Should probably be informed. Should the target be informed?

You mean the one that effect is redirected onto?

venomousbirds:
> > Should probably be informed. Should the target be informed?
> 
> You mean the one that effect is redirected onto?

I was thinking the one that targets the assistant, but that makes no sense, nor does the target that the effect is redirected onto

Shepherdwhen:
Should infection be immune to redirection by assistant to prevent a scenario where the assistant redirects an infected wolf onto a wolf in the pack or infected wolf itself

venomousbirds:
I think that's a good thing, if it gets redirected. Keeps wolves from just randomly infecting someone

McTsts:
> is the coward informed when they redirect something? it doesnt mention it in the description

Added a thing so that the Assistant is informed when they redirect something, but not what they redirected

PR #166, opened 2021-04-22
Create innkeeper

McTsts:
the role that was discussed on MW


Mjnstag:
Would the inn be deleted/disbanded when the inkeeper dies?

McTsts:
> Would the inn be deleted/disbanded when the inkeeper dies?

Hmm, good question. Probably should be?

Mjnstag:
it probably should i think.

Shepherdwhen:
Does that line mean that 3 flute players could enchant 3 people with 20 players?

Shepherdwhen:
Wrong issue :s

McTsts:
Changed that!

McTsts:
This wasn't used all that much, but I don't really see anything that could go wrong with this. Doesn't really add anything new.

venomousbirds (approved):

Issue #155, opened 2021-04-19
Fox Suggestion


McTsts:
Maybe we should make Fox abilities end-day abilities so that they aren't immediate and so that they can switch to joining the pack instead?

Shepherdwhen:
What if the CF can ask for it to be immediate of EoD each time because an immediate is more useful D0 but if the CF gets later into the game if it's not found the pack it's more likely to die so may choose an EoD check

MrTurtleJr:
> 
> 
> What if the CF can ask for it to be immediate of EoD each time because an immediate is more useful D0 but if the CF gets later into the game if it's not found the pack it's more likely to die so may choose an EoD check
Doesn't that over-complicate things? As far as I'm concerned, no other roles can do this, so why the exception? I'd be in support of an end-day ability.

McTsts:
Yeah I agree, that seems a bit too over-complicated

Shepherdwhen:
> > What if the CF can ask for it to be immediate of EoD each time because an immediate is more useful D0 but if the CF gets later into the game if it's not found the pack it's more likely to die so may choose an EoD check
> 
> Doesn't that over-complicate things? As far as I'm concerned, no other roles can do this, so why the exception? I'd be in support of an end-day ability.

Maybe, just feels like the D0 check is more useful being immediate as you're either gonna find a wolf and not die or find a good kill target for wolves, after that an End of Day check is much better

MrTurtleJr:
Yeah, but I don't know if we should cater to D0 for the ability, as the ability will more often than not be used on a non-D0 day. I also just think that consistency would make already complicated roles easier to understand (re: new(er) players).

McTsts:
> > > What if the CF can ask for it to be immediate of EoD each time because an immediate is more useful D0 but if the CF gets later into the game if it's not found the pack it's more likely to die so may choose an EoD check
> > 
> > 
> > Doesn't that over-complicate things? As far as I'm concerned, no other roles can do this, so why the exception? I'd be in support of an end-day ability.
> 
> Maybe, just feels like the D0 check is more useful being immediate as you're either gonna find a wolf and not die or find a good kill target for wolves, after that an End of Day check is much better

the wolves only get the info EoD anyway, but I was actually considering suggestion that DF joins the pack if they find a wolf

Shepherdwhen:
[...]

PR #172, opened 2021-04-23
Update clairvoyant fox

McTsts:
As discussed in #155
I also added a thing so pack cant kill them


Shepherdwhen:
> As discussed in #155
> I also added a thing so pack cant kill them

That seems like they have no reason to join pack since wolves can't really kill them so they will stay CF for ages

Shepherdwhen:
> > As discussed in #155
> > I also added a thing so pack cant kill them
> 
> That seems like they have no reason to join pack since wolves can't really kill them so they will stay CF for ages

I didn't read it fully no it works lol

McTsts:
hmm, possibly, but on the other hand as long as they don't join their info is also worthless

Shepherdwhen:
> hmm, possibly, but on the other hand as long as they don't join their info is also worthless

I just misread the part about forced wolf so I thought it was just like a runner not a cc

PR #181, opened 2021-04-25
Update mad scientist

### LINKED ISSUE #174: Is Mad Scientist revealed
Are they revealed on death?

McTsts:
Fixes #174


Mjnstag:
so the MS doesnt get revealed when they hit a scared wolf for example since they only get revealed when they kill someone? otherwise change it to attacks?

McTsts:
> so the MS doesnt get revealed when they hit a scared wolf for example since they only get revealed when they kill someone? otherwise maybe change it to attacks?

Honestly, I thought that's also was how Huntress worked, apparently it is not?

venomousbirds:
I think same as I think about huntress: if the target survives, the huntress doesn't get revealed. mad scientist shouldn't get revealed either in the same case

PR #185, opened 2021-04-27
Create Mirror and Shiny Wolf

bobloy:
Power reflection abilities, for town and wolves


McTsts:
The bad luck part seems kinda out of place in Mirror imo (and also seems too much like Angel/Despot)

venomousbirds:
this is like some redirection role with a despot flair. not a big fan of redirection, and the despot thing should probably stay for despot only

bobloy:
Removed the Mirror's curse, reduced just to a redirection ability.

McTsts:
I think for now we don't want to add more redirections.

PR #186, opened 2021-04-27
Create Frankenstein's Monster

bobloy:
Adds Frankenstein's monster, a reporter-like power that gathers abilities from the dead.


Shepherdwhen:
This sounds like town version of reaper, and reaper sort of only just works as a solo

McTsts:
This could just replace Reaper 
hepi

Shepherdwhen:
It can be strong actually as it's a confirmed town reporter
!thonk

Shepherdwhen:
how do you use emotes?

Mjnstag:
> how do you use emotes?

upload an image in the comment
image

venomousbirds:
This sounds more like reaper than a town role. I don't think a town role should be this powerful

McTsts:
> This could just replace Reaper

This PR inspired me to make a new Reaper (an unfinished solo role) version!
I've combined this suggestion with it for the new version. I think this will work better there than as a townsfolk role. You can find that role under #283.

I'll close this because I think we're more likely to use #283 than this one, but thanks for giving us that idea!

PR #187, opened 2021-04-27
Create Teenage Wolf

bobloy:
Basic wolf with a downside, requiring the wolf to use the word "wolf" at least once each day in the town square.

### DIFF
--- a/limited/teenage wolf
+++ b/limited/teenage wolf
@@ -1,0 +1,5 @@
+**Teenage Wolf** | Werewolf Miscellaneous
+__Basics__
+The Teenage Wolf must use the word "wolf" at least once in the town square each day, or will they will die at the end of the day.
+__Details__
+Variations of "wolf" are allowed, such as: wolves, werewolf, werewolves.


Shepherdwhen:
Definitely really complex, plus inactivity basically is a reason why roles that requires speech never work

bobloy:
Discord search makes it pretty easy to verify, but I agree that inactivity causes this role to die. Not sure if that's a bad thing

McTsts:
I don't think this is a bad role per see (and it probably works in the places where it's being used), but I don't think this is really compatible with our games. I think this would just be very annoying to deal with as you'd need to keep track of what everyone says on town square, which can sometimes be a lot. I definitely think we should not add this.

venomousbirds (commented):
I really dislike this role, because we can't monitor the chat all day, people will simply meta it by not talking/find some meta to easily find the wolf and it does not fit werewolves at all. Dying because of not saying a specific word in a chat is a horrible way to go.

bobloy:
Thanks for the feedback. I think it's safe to close this PR based on the discussion.

PR #188, opened 2021-04-27
Create idiot

McTsts:
idiot


venomousbirds:
idiot

Mjnstag:
idiot
if you add this, maybe it would be a good idea to note the way the role works (after losing a vote) in relation to vote stealing roles.

venomousbirds:
I still think idiot should be told when their vote is taken away

McTsts:
> I still think idiot should be told when their vote is taken away

I already added that.

McTsts:
> idiot
> if you add this, maybe it would be a good idea to note the way the role works (after losing a vote) in relation to vote stealing roles.

Good point.

venomousbirds:
> > I still think idiot should be told when their vote is taken away
> 
> I already added that.

Oops, I thought you just added the old idiot description that didn't inform the vote being taken away

McTsts:
> > > I still think idiot should be told when their vote is taken away
> > 
> > 
> > I already added that.
> 
> Oops, I thought you just added the old idiot description that didn't inform the vote being taken away

Nope, this one is changed a bit.

bobloy:
Does an idiot mayor have zero votes or one vote? Mayor grants an "extra vote"

McTsts:
> Does an idiot mayor have zero votes or one vote? Mayor grants an "extra vote"

Interesting question. From a bot functionality perspective we'd have to decide between it having the extra vote (the Mayor vote always adds +1 or -1 votes to your vote value) or the attacked Idiot not having access to #voting-booth (which would reveal them as attacked idiot)

Shepherdwhen:
> > Does an idiot mayor have zero votes or one vote? Mayor grants an "extra vote"
> 
> Interesting question. From a bot functionality perspective we'd have to decide between it having the extra vote (the Mayor vote always adds +1 or -1 votes to your vote value) or the attacked Idiot not having access to #voting-booth (which would reveal them as attacked idiot)

What if we made an identical mayor role that just doesn't have an extra vote or is that a dumb fix?

McTsts:
> > > Does an idiot mayor have zero votes or one vote? Mayor grants an "extra vote"
> > 
> > 
> > Interesting question. From a bot functionality perspective we'd have to decide between it having the extra vote (the Mayor vote always adds +1 or -1 votes to your vote value) or the attacked Idiot not having access to #voting-booth (which would reveal them as attacked idiot)
> 
> What if we made an identical mayor role that just doesn't have an extra vote or is that a dumb fix?

It's still possible to figure out the difference from those roles

Shepherdwhen:
> > > > Does an idiot mayor have zero votes or one vote? Mayor grants an "extra vote"
> > > 
> > > 
> > > Interesting question. From a bot functionality perspective we'd have to decide between it having the extra vote (the Mayor vote always adds +1 or -1 votes to your vote value) or the attacked Idiot not having access to #voting-booth (which would reveal them as attacked idiot)
> > 
> > 
> > What if we made an identical mayor role that just doesn't have an extra vote or is that a dumb fix?
> 
> It's still possible to figure out the difference from those roles

Oh yeah, tbh the only real way would be +1 vote, orrrrr a change in mayor so that if a mayor player is idiot they get +0 once attacked or or when you use the command to remove vote power you remove 2 and then mayor fills the gap of the -1 vote making it 0

McTsts:
> > > > > Does an idiot mayor have zero votes or one vote? Mayor grants an "extra vote"
> > > > 
> > > > 
> > > > Interesting question. From a bot functionality perspective we'd have to decide between it having the extra vote (the Mayor vote always adds +1 or -1 votes to your vote value) or the attacked Idiot not having access to #voting-booth (which would reveal them as attacked idiot)
> > > 
> > > 
> > > What if we made an identical mayor role that just doesn't have an extra vote or is that a dumb fix?
> > 
> > 
> > It's still possible to figure out the difference from those roles
> 
> Oh yeah, tbh the only real way would be +1 vote, orrrrr a change in mayor so that if a mayor player is idiot they get +0 once attacked or or when you use the command to remove vote power you remove 2 and then mayor fills the gap of the -1 vote making it 0

if you have -1 votes with mayor it's -2

Shepherdwhen:
> > > > > > Does an idiot mayor have zero votes or one vote? Mayor grants an "extra vote"
> > > > > 
> > > > > 
> > > > > Interesting question. From a bot functionality perspective we'd have to decide between it having the extra vote (the Mayor vote always adds +1 or -1 votes to your vote value) or the attacked Idiot not having access to #voting-booth (which would reveal them as attacked idiot)
> > > > 
> > > > 
> > > > What if we made an identical mayor role that just doesn't have an extra vote or is that a dumb fix?
> > > 
> > > 
> > > It's still possible to figure out the difference from those roles
> > 
> > 
> > Oh yeah, tbh the only real way would be +1 vote, orrrrr a change in mayor so that if a mayor player is idiot they get +0 once attacked or or when you use the command to remove vote power you remove 2 and then mayor fills the gap of the -1 vote making it 0
> 
> if you have -1 votes with mayor it's -2

Oh yeah Idk then mayor + idiot is gonna be weird to do

bobloy:
Sounds like there needs to be a multiplier step for voting power. Idiot would change multiplier from default 1 to 0.

Mjnstag:
what would be the message to wolves/other attackers when the target doesnt die? "nothing happened"?

venomousbirds (approved):

McTsts:
> what would be the message to wolves/other attackers when the target doesnt die? "nothing happened"?

Yeah, just generic failure, like Devil/Scared Wolf/Redirection/Protected

PR #206, opened 2021-04-29
Major Devil Revamp

McTsts:
- changed the invest option (now you need to sell your identity) (wager)
- clarified that ignoring is an option (wager)
- added a protection option (wager)
- the devil now refuses to kill anyone from hell team and not just the devil
- clarify that auto soul defense uses 1 soul per attack
- split wager and soul info into different sections
- demon renamed to imp (only meant for copying)
- demon readded as a role that's like devil but without the wager (which is meant to be added to role lists)
- the souls are now in a pool of souls and not on the devil
- added separate more detailed info for wager & souls
- each Devil and Demon adds 1 soul to the initial pool of souls
- when an imp is created hell team gains a soul

### DIFF
--- a/channels/wager
+++ b/channels/wager
@@ -2,8 +2,10 @@
 
 You can decide to either:
 
-1. Kill someone, and sell your soul in the process
+1. Kill someone, but sell your soul in the process
 
-2. Check someone's role, but the Devil will also see the result.
+2. Protect someone, but sell your soul in the process
 
-3. Ignore, and don't do anything.
+3. Check someone's role, but the Devil will see the result and also your role
+
+4. Ignore, and don't do anything.
--- a/channels/wager
+++ b/channels/wager
@@ -2,9 +2,9 @@
 
 You can decide to either:
 
-1. Kill someone, but sell your soul in the process
+1. Attack a player, but sell your soul in the process
 
-2. Protect someone, but sell your soul in the process
+2. Protect yourself or any player, but sell your soul in the process
 
 3. Check someone's role, but the Devil will see the result and also your role
 
--- a/other/hell/demon
+++ b/other/hell/demon
@@ -7,7 +7,7 @@ At the start of the game, the Demon may choose a role as a weak disguise. If the
 
 The Demon is soulless and their soul is in the hell team's pool of souls.
 The Demon may use souls from the pool to protect or attack somebody. 
-If the Demon is attacked while a soul is in the pool the soul will automatically be used to protect the Demon. 
+If the Demon is attacked while a soul is in the pool, a soul will automatically be used to protect the Demon. 
 Use `$info souls` for a detailed explanation.
 
 Members of the Hell team cannot change role or betray the team in any way.
--- a/other/hell/demon
+++ b/other/hell/demon
@@ -5,7 +5,7 @@ Just like the devil the Demon can use the hell team's pool of souls.
 __Details__
 At the start of the game, the Demon may choose a role as a weak disguise. If the Demon doesn’t choose a role, they will show up as a citizen.
 
-The Demon is soulless and their soul is in the hell team's pool of souls.
+At the start of the game the Demon's soul is added to the hell team's pool of souls, and they are soulless afterwards.
 The Demon may use souls from the pool to protect or attack somebody. 
 If the Demon is attacked while a soul is in the pool, a soul will automatically be used to protect the Demon. 
 Use `$info souls` for a detailed explanation.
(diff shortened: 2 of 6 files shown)


channels/wager
```diff
@@ -2,8 +2,10 @@
 
 You can decide to either:
 
-1. Kill someone, and sell your soul in the process
+1. Kill someone, but sell your soul in the process
 
-2. Check someone's role, but the Devil will also see the result.
+2. Protect someone, but sell your soul in the process
```
Mjnstag, line 7: maybe include a "(including yourself)" since it you can also protect yourself and it kinda sounds like you cant.

other/hell/demon
```diff
@@ -1,8 +1,13 @@
...
+**Demon** | Solo Killing - Hell Team
 __Basics__
 The Demon wins with the devil and joins a channel with the devil to communicate securely.
+Just like the devil the Demon can use the hell team's pool of souls.
 __Details__
-The Demon is usually the result of someone copying the devil for various reasons. This is because the devil is too powerful for there to be two. 
-Their win condition is the same as the Devil’s. The Demon starts out soulless.
+At the start of the game, the Demon may choose a role as a weak disguise. If the Demon doesn’t choose a role, they will show up as a citizen.
+
+The Demon is soulless and their soul is in the hell team's pool of souls.
```
Mjnstag, line 8: maybe mention the starting pool of souls?

other/hell/devil
```diff
@@ -1,22 +1,21 @@
...
-
-If the player chooses to see another player’s role (immediate), it will be affected by weak disguises and obstructions. These effects are seen from the player’s perspective, not the Devil’s.
-If the player chooses to attack another player (soul sold: immediately; kill executed: end-night), then they will become soulless. The attacker is the Devil. 
-If the player chooses to attack the Devil the attack will not be executed, but they will not be informed of this. However, they still sell their soul to the Devil.
+At the start of the game, the Devil may choose a role as a weak disguise. If the Devil doesn’t choose a role, they will show up as a citizen.
 
-The Devil can use a soul to protect someone else from all attacks for the next night (day; immediate) or kill a player of their liking (day/night; immediate). 
-If the Devil is attacked at night and they have spare souls, they will be protected and a soul will be used up automatically. 
+The Devil is soulless and their soul is in the hell team's pool of souls.
```
Mjnstag, line 10: same as demon, maybe mention it affects the starting amount?

other/hell/souls
```diff
@@ -0,0 +1,10 @@
+**Souls** | Additional Info - Hell Team
+Members of the hell team that can use Souls may use a Soul to protect someone from all attacks for the current night. If used during the day the player will be protected for the next night.
+Using a Soul to protect a player is an immediate effect.
+
+Members of the hell team that can use Souls may use a Soul to kill a player. The player will be killed immediately and it is not revealed why they died.
```
Mjnstag, line 5: It mentions attacking someone when they sell their soul but here it mentions killing someone. i think it should be the former? Also is it a good idea to mention a failing message if it indeed is the former?

channels/wager
Mjnstag, line 7: since when hell team gets attacked, it uses a soul for every attack but wager protecting doesnt?. Imo it feel a bit weird but i get that otherwise that option might not seem as interesting. 

Soul info: "If they are attacked several times a soul is used for every attack. "

other/hell/demon
McTsts: Did you miss this?

> The Demon is soulless and their soul is in the hell team's pool of souls.

Alternatively I might be misunderstanding what you're saying

other/hell/devil
McTsts: Same as Demon

> The Devil is soulless and their soul is in the hell team's pool of souls.

other/hell/souls
McTsts: Yeah it should be attacking. There is no failing message for the recipient though.

other/hell/demon
Mjnstag: they are not in there the whole game if they get used

other/hell/souls
McTsts: Misread for what this is. Ignore the part about recipient.

other/hell/demon
McTsts: Ah, I see.

channels/wager
McTsts: That's how it always was.

Mjnstag:
👍 looks good

PR #207, opened 2021-04-29
fixes #205

### LINKED ISSUE #205: Tanner Disguise
clarify that the Tanner is not affected by strong disguises.

Mjnstag:
might need to change some sentences around so the beginning of the last 2 sentences are not the same.


McTsts:
combining those two should be fine

PR #209, opened 2021-04-30
Create Juror

voidmist:
This is a group/investigative role that I think can work well in big games. They're essentially bakers who can investigate every other day but have the risk of disbanding if enough of they die/become undead. I'm not sure if this falls in the group category or the investigative category.

_To note_: The jury is affected by flute player obstructions as long as at-least one juror is enchanted. 
_To clarify_: Only the **role** is announced publicly not the player investigated.

### DIFF
--- a/townsfolk/investigative/Juror
+++ b/townsfolk/investigative/Juror
@@ -1,0 +1,9 @@
+**Juror** | Townsfolk Investigative/Group
+__Basics__
+The Jurors have a secret channel where they can communicate among themselves securely. On alternate days, the Jury votes to arraign and investigate a player.
+If the Jury successfuly arraigns a non-town role, it is announced.
+__Details__
+The Juror is part of the Jury. After Day 0, on odd days, the cult can arraign a player, investigating them. 
+The Jury recieves the result of the investigation at the end of the day. If the result of the investigation shows a non-town role, the role is immediately announced publicly.
+The Jury is disbanded if there are 1 or more Jurors left; this means the Jury is active only if atleast two Jurors are alive. 
+Any member of the Jury, even if they lose their role, can still vote on who to arraign. The Jury is affected by weak disguises and obstructions.


townsfolk/investigative/Juror
```diff
@@ -0,0 +1,9 @@
+**Juror** | Townsfolk Investigative/Group
+__Basics__
+The Jurors have a secret channel where they can communicate among themselves securely. On alternate days, the Jury votes to arraign and investigate a player.
+If the Jury successfuly arraigns a non-town role, it is announced.
+__Details__
+The Juror is part of the Jury. After Day 0, on odd days, the cult can arraign a player, investigating them. 
+The Jury recieves the result of the investigation at the end of the day. If the result of the investigation shows a non-town role, the role is immediately announced publicly.
+The Jury is disbanded if there are 1 or more Jurors left; this means the Jury is active only if atleast two Jurors are alive.
```
McTsts, line 8: Do you mean one or less?

townsfolk/investigative/Juror
```diff
@@ -0,0 +1,9 @@
+**Juror** | Townsfolk Investigative/Group
+__Basics__
+The Jurors have a secret channel where they can communicate among themselves securely. On alternate days, the Jury votes to arraign and investigate a player.
```
McTsts, line 3: probably should be odd instead of alternate

townsfolk/investigative/Juror
```diff
@@ -0,0 +1,9 @@
...
+If the Jury successfuly arraigns a non-town role, it is announced.
```
McTsts, line 4: From the way it's written it seems like "arraign" has some sort of relevant meaning, but from what I can tell it doesn't and it's just normal investigation?

townsfolk/investigative/Juror
```diff
@@ -0,0 +1,9 @@
+**Juror** | Townsfolk Investigative/Group
```
McTsts, line 1: Hmm, I think Townsfolk Group would probably make more sense here (it can't be both).
I know we usually do TK > TI > TP > TM, but I'm not sure if we have a consistent place for TG. I'd say we should probably do TK > TG > TI >TP > TM which would be consistent with what we have right now and would mean that this role is a TG role.

McTsts (commented):
This seems like a fairly interesting idea I think.

townsfolk/investigative/Juror
voidmist: Sorry yeah, 1 or less.

townsfolk/investigative/Juror
```diff
@@ -0,0 +1,9 @@
...
+__Basics__
+The Jurors have a secret channel where they can communicate among themselves securely. On alternate days, the Jury votes to arraign and investigate a player.
+If the Jury successfuly arraigns a non-town role, it is announced.
+__Details__
+The Juror is part of the Jury. After Day 0, on odd days, the cult can arraign a player, investigating them. 
+The Jury recieves the result of the investigation at the end of the day. If the result of the investigation shows a non-town role, the role is immediately announced publicly.
+The Jury is disbanded if there are 1 or more Jurors left; this means the Jury is active only if atleast two Jurors are alive. 
+Any member of the Jury, even if they lose their role, can still vote on who to arraign. The Jury is affected by weak disguises and obstructions.
```
McTsts, line 9: For consistency with other groups like wolves/cult I'd suggest the following:
The jurors get a poll just like the cult. One random juror that voted for the winner of the poll is chosen to execute the investigation. The obstruction is only applied if it applies to that Juror specifically.

townsfolk/investigative/Juror
voidmist: yeah you're right. I think I was overcomplicating things with an unnecessary word. Its just an investigation.

townsfolk/investigative/Juror
voidmist: Yeah that works

townsfolk/investigative/Juror
Shepherdwhen: What if we did it like cult
As long as like a head juror lives then they can investigate
If they die/don't exist it's just a TG but doesn't disband
And juror is just a TG

voidmist: That's a decent solution but I didn't want to make it too similar too cult. It would be more unique if its vulnerability lies in the numbers and not a singular role.

McTsts: I agree. I think the CL thing should be unique to Cult.

townsfolk/investigative/juror
```diff
@@ -0,0 +1,9 @@
+**Juror** | Townsfolk Group
+__Basics__
+The Jurors have a secret channel where they can communicate among themselves securely. On alternate days, the Jury votes to investigate a player.
```
McTsts, line 3: This should be "odd" instead of "alternate"

townsfolk/investigative/juror
```diff
@@ -0,0 +1,9 @@
...
+If the Jury successfuly investigates a non-town role, it is announced.
```
venomousbirds, line 4: Not a fan of just announcing a role, even if it's only a role and not a complete check. I like the idea of the Jurors but this is not really neccesary

venomousbirds (changes requested):

townsfolk/investigative/juror
McTsts: Yeah, actually I think I agree. I think this could still work without the announcement.

voidmist: I felt like the announcing mechanic was interesting because it lets all parties have an understanding of where the game is and shift the game as a whole.

Its an obvious advantage to town as:
1.  Townies know that said role has been confirmed and can use that info accordingly
2.  A juror/jury can effectively confirm themselves without necessarily revealing everyone in the jury
3.  Makes way for an interesting mechanic where you have discussion in town hall on who the jury would check. Obviously
there's no guarantee the Jury would've followed through with public requests.

Its advantageous to wolves since:
1.  Town can no longer hide the fact that they know a person is evil, which means they cant _pretend_ to trust someone to keep their faction under a false sense of security.
2.  Revealed roles will have to change their gameplay, becoming bait/sacrifice or becoming an ideal member to go scouting for non-pack members.

Also, the jury would have to play differently and uniquely than other investigative roles and would really be involved with public census. They have to choose carefully as to not cause panic, confusion and tip off the wolves while also trying not to out themselves. Further, in a more meta sense, town-hall activity would increase and bring people out of conspiracy channels.
Removing the announcement thing just makes them fortune teller bakers in all honesty.

McTsts: You're brought up a bunch of things I didn't think of. I'm convinced.

McTsts (approved):

townsfolk/investigative/juror
```diff
@@ -0,0 +1,11 @@
+**Juror** | Townsfolk Group
+__Basics__
+The Jurors have a secret channel where they can communicate among themselves securely. On odd days, the jury votes to investigate a player.
+If the jury successfuly investigates a non-town role, it is announced.
+__Details__
+The Juror is part of the jury. After Day 0, on odd days, the jury can vote to investigate a player. 
+The jury recieves the result of the investigation at the end of the day. If the result of the investigation shows a non-town role, the role is immediately announced publicly.
+The jury is disbanded if there are 1 or less Jurors left; this means the jury is active only if atleast two Jurors are alive. 
+Any member of the jury, even if they lose their role, can still vote on who to investigate. 
+One of the Jurors that has voted for the jury's target is chosen as the investigator, who performs the investigations.
```
Mjnstag, line 10: should it be added that it is randomly chosen from the ones who voted or can the jury choose? how is this impacted when a juror has lost their role but gets chosen to investegate, since imo they cant investiage anymore since they arent that role anymore.

McTsts: Oh yeah, I forgot to add random.

McTsts: think for consistency Jurors that lost their role would still be able to vote and investigate.

Mjnstag: yea, probably the for the best

PR #212, opened 2021-04-30
Remove Ice King

McTsts:
at least for now


venomousbirds (approved):
ice king has potential but i dont know how to change it to make it better and until that happens, i agree with this

McTsts:
> ice king has potential but i dont know how to change it to make it better and until that happens, i agree with this

Yeah, exactly what I also think

Mjnstag (approved):

PR #219, opened 2021-05-02
Update aura teller

### LINKED ISSUE #214: Aura Teller redirection immunity
AT is immune to redirections

McTsts:
Fixes #214


McTsts:
Pretty sure this is how it's supposed to be

Shepherdwhen:
It makes the Fpa harder to play which I think works, it means an FPA targeting a wolf can't be checked as not a threat by AT

venomousbirds (approved):

PR #222, opened 2021-05-03
Update pack tanner

### LINKED ISSUE #194: Pack Tanner nerf
[VIP]  pack tanner is probably too strong even with just 1 disguise - possible solution: it can only tan players not in the pack

McTsts:
Fixes #194


venomousbirds (approved):
Good consistent way to nerf roles

PR #260, opened 2021-05-05
Update acrobat

Mjnstag:
Changed description so that there are less differences between runner and acrobat description


McTsts (approved):

Shepherdwhen:
Someone suggested acrobat should survive 1 normal kill and 2 wolf kills? or did we say that was a bad idea?

Shepherdwhen:
It was in a minis game once

McTsts:
> Someone suggested acrobat should survive 1 normal kill and 2 wolf kills? or did we say that was a bad idea?

Interesting suggestion! You can make it a different issue or PR if you want.

PR #268, opened 2021-05-06
Add Tasty Villager

bobloy:
A role used to balance town favored matches.


McTsts:
Macho actually can't be killed no matter how many times you attack them.

bobloy:
> Macho actually can't be killed no matter how many times you attack them.

Good to know, will remove Macho as an example.

Shepherdwhen:
> I think a town role that balance in the wolves favor would be nice to have. I think right now we only have CC which is a very different playstyle.

I think another suggestion for one of these wolf favouring town roles was a town who is strongly disguised as wolf, possibly we should draft a PR for it

venomousbirds (changes requested):
I don't think this role is useful, or fun to play. Your goal is to stay alive, yes, but town that favours wolves is not something I think we need in Werewolves. There's Wolf's Cub that practically does the same

bobloy:
> I don't think this role is useful, or fun to play. Your goal is to stay alive, yes, but town that favours wolves is not something I think we need in Werewolves. There's Wolf's Cub that practically does the same

I disagree on it being useful. To use a 5 player game as an example: wolf cub would do nothing as a solo wolf, so that type of ability could only work on a town role.

As for it being fun to play, I think it's clearly more fun that citizen, and probably more fun than bard.

McTsts:
> I don't think this role is useful, or fun to play. Your goal is to stay alive, yes, but town that favours wolves is not something I think we need in Werewolves. There's Wolf's Cub that practically does the same

CC is another wolfsided town role. This one could also be a nice addition to the Citizen/Journalist type of role category. This role could probably lead to an interesting playstyle, more so than CC as there's no risk of betrayal.

McTsts:
> > I don't think this role is useful, or fun to play. Your goal is to stay alive, yes, but town that favours wolves is not something I think we need in Werewolves. There's Wolf's Cub that practically does the same
> 
> I disagree on it being useful. To use a 5 player game as an example: wolf cub would do nothing as a solo wolf, so that type of ability could only work on a town role.
> 
> As for it being fun to play, I think it's clearly more fun that citizen, and probably more fun than bard.

Semi-unrelated, but Bard is actually a fairly powerful role and can not be used like other weak TM's (Journalist, Citizen, this, maybe Child)

Shepherdwhen:
> > I don't think this role is useful, or fun to play. Your goal is to stay alive, yes, but town that favours wolves is not something I think we need in Werewolves. There's Wolf's Cub that practically does the same
> 
> CC is another wolfsided town role. This one could also be a nice addition to the Citizen/Journalist type of role category. This role could probably lead to an interesting playstyle, more so than CC as there's no risk of betrayal.

CC is also very likely to be lynched if there is no attack and no runner claims, this is very unlikely to be lynched

bobloy:
Some name suggestions other than Fat Guy:

Tasty Citizen
Catalyst
Accelerant
Sacrificial Lamb / Lamb
Buffet

McTsts:
maybe Tasty Villager, because right now we don't do variants of Citizen

McTsts:

PR #269, opened 2021-05-06
Update & Rebrand Evil Scientist

### LINKED ISSUE #223: Possible Evil Scientist Revamp
- players are instantly contagious
- remove death reason announcement
- rewrite

McTsts:
Fixes #223


Mjnstag:
> Their desire is to kill off all players.
maybe "all other players" since it it sounds now that they have to die as well to win

Shepherdwhen:
I was thinking that the contagious disease being immediate night leads to the problem that immediate abilities before the player is contaminated won't apply, e.g

Hooker targets runner
Evil scientist targets the same runner
Aura teller targets the same runner

In this scenario hooker is not contaminated but aura teller is
The Evil scientist should be an end day ability really to avoid this happening

McTsts:
This PR should also change the name of the role

McTsts:
A new name for Lab Assistant (and possibly Immune Survivor) is required.

Shepherdwhen:
We could rename lab assistant to plague carrier and rename immune survivor to Plague Doctor

McTsts:
that could work

PR #271, opened 2021-05-06
Acrobat should have same escape message as Runner

### LINKED ISSUE #227: Acrobat & Runner
Are wolves supposed to be able to differentiate between acrobat and runner?

McTsts:
Fixes #227


venomousbirds (commented):
Runners just escape, as do acrobats. I've never seen people use anything different for acrobat and runner

McTsts:
> Runners just escape, as do acrobats. I've never seen people use anything different for acrobat and runner

> If an Acrobat evades a kill from a wolf outside of the pack, the wolf responsible is told why.

Makes it sound like they're told that an _Acrobat_ evaded the attack, this PR makes it so it's what you said.

venomousbirds:
I don't think writing an extra sentence does much for the role, except for having people to look up runner again. Maybe mentioning "When the Runner/Acrobat escapes, the wolves are only notified that their target escaped" in both descriptions works better?

Dennis-0 (approved):

Dennis-0:
I assumed people will know what a runner is, but if we think some people might need to lookup runner in that case, then I think Vera's line is better

PR #276, opened 2021-05-07
Executioner Clarification

McTsts:
Not sure if this wording is particularly good..

### DIFF
--- a/townsfolk/killing/executioner
+++ b/townsfolk/killing/executioner
@@ -2,7 +2,7 @@
 __Basics__
 Whenever the Executioner is chosen to be lynched, it will be prevented, and a player chosen by the Executioner will be lynched in their place.
 __Details__
-The Executioner may target a player. This target can be updated once per phase, with 1 emergency change to be used whenever the Executioner wants to. Targetting a player is an immediate effect.
-This player will be lynched instead of the Executioner. The Executioner may alternatively choose to not target any player. 
+The Executioner may target a player. This target can be updated once per phase, with 1 emergency change to be used whenever the Executioner wants to. Targeting a player is an immediate effect. The Executioner may alternatively choose to not target any player. 
+If the Executioner would be lynched, their target is lynched instead. If there is no target, nobody is lynched. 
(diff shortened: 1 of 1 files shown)


townsfolk/killing/executioner
```diff
@@ -4,5 +4,5 @@ Whenever the Executioner is chosen to be lynched, it will be prevented, and a pl
 __Details__
 The Executioner may target a player. This target can be updated once per phase, with 1 emergency change to be used whenever the Executioner wants to. Targetting a player is an immediate effect.
 This player will be lynched instead of the Executioner. The Executioner may alternatively choose to not target any player.
```
bobloy, line 6:
```suggestion
The Executioner may target a player. This target can be updated once per phase, with 1 emergency change to be used whenever the Executioner wants to. Targeting a player is an immediate effect. The Executioner may alternatively choose to not target any player. 
If the Executioner would be lynched, their target is lynched instead. If there is no target, the Executioner is lynched as normal. 
```

townsfolk/killing/executioner
```diff
@@ -4,5 +4,5 @@ Whenever the Executioner is chosen to be lynched, it will be prevented, and a pl
 __Details__
 The Executioner may target a player. This target can be updated once per phase, with 1 emergency change to be used whenever the Executioner wants to. Targetting a player is an immediate effect.
 This player will be lynched instead of the Executioner. The Executioner may alternatively choose to not target any player. 
-When the lynch fails, the Executioner’s role is revealed to all, and the targeted player is lynched immediately. 
-After the failed lynch, the Executioner, can no longer use their ability.
+When the Executioner changes the lynch, their role is revealed to all, and the targeted player is lynched immediately. 
+After that, the Executioner, can no longer use their ability.
```
bobloy, line 8:
```suggestion
When the Executioner changes the lynch, their role is revealed to all, and the targeted player is lynched immediately. 
This ability will only work once. Future lynches will behave as normal.
```

bobloy (changes requested):
Some wording suggestions.

townsfolk/killing/executioner
bobloy: Updated

PR #304, opened 2021-05-10
Create Survivor

voidmist:
An interesting limited, unaligned role that kinda goes against the conventions of unaligned but this can still work in particular games.


venomousbirds:
Unaligned roles don't take the win, they can win with anyone. All other alignments can win with the unaligned still alive. This role has the win condition of a Solo role.

voidmist:
I don't think this is a good role in hindsight, so I'm closing it.

PR #309, opened 2021-05-12
Create direwolf

McTsts:
A suggestion for another wolf role for variety. For this one the pack would have to be extra careful that they don't die, could lead to some interesting playstyles.


Shepherdwhen:
So reverse wolf cub?
As long as it's alive and no one is wolves get an extra kill

McTsts:
Kinda of, I guess.

McTsts:
This one of the ones that was suggested a while ago and discussed by the auditors.

venomousbirds:
might be a little powerful in smaller games where wolves are low in numbers? (wolves might die faster and direwolf might be powerful then)

Shepherdwhen:
Imagine direwolf in minis

McTsts:
This is feedback it got in auditors:

Chopper:
>  again sounds like a very niche/rare thing to occur, but another semi-useless wolf role is fine I guess. I'm worried that the direwolf itself can just change the game too much on its own pretty much based on random chance. For example if the direwolf dies early, then nothing happens and it's just like he/she was another wolf. But if he/she just so happens to not be killed until the end, then the wolf team suddenly has a massive advantage. Generally people don't know exactly what type of wolf a person is when they choose to kill it. This might just feel like a "wtf, we were doing so well and then this randomly came out of nowhere"

WhyThough:
> good if late-game comeback is possible

McTsts:
I think some of those points _can_ be accurate, but it doesn't have to be. The pack can play in a way to keep the Direwolf save (possibly with the help of Pack Tanner or Tanner) even, which would definitely be a noticeable difference and not just a random chance.

McTsts:
> > intersting counter to Exile
> 
> so whats exile?

It was another of the suggestions, but I didn't think it was very good.

> Once in a game, the Exile may ask for the count of werewolf team members left in the game. If the number of living wolf roles is greater or equal to the town roles, the Exile may kill a player each night.

venomousbirds:
that's very luck based and very weird

McTsts:
Yeah, I'm not a fan of it.

McTsts:
I like Direwolf because the pack will know and can base claims/strategies off it, but Exile doesn't have anything like that.

venomousbirds:
äs a counter to exile", but thats weird to say then, how is it a counter to exile

McTsts:
> "as a counter to exile", but thats weird to say then, how is it a counter to exile

Yeah idk. I think it's fine on it's own.

McTsts:

PR #310, opened 2021-05-12
Create Alert Wolf

Shepherdwhen:
An Idea that came from a misunderstanding of Sacred/Blessed wolf and is a slightly buffed version of scared wolf that knows what it was attacked by


McTsts:
Merged with #126

PR #311, opened 2021-05-12
Update infected wolf

Shepherdwhen:
Just some SPAG improvements and We never changed IW to becoming imp it still said Demon


McTsts:
A current style choice is that only the role name of the current role is capitalized
>  In a role's description the name of the role itself is capitalized, but other role names are not.

Shepherdwhen:
> A current style choice is that only the role name of the current role is capitalized
> 
> > In a role's description the name of the role itself is capitalized, but other role names are not.

This one was 2/3 capitalised in the description so I went with the majority which was capitalised.

PR #326, opened 2021-05-18
Create Alchemist

12Metatron:
New role idea. Unaligned Miscelaneus


McTsts:
Sounds interesting in theory, but I'm not sure if people would really be willing to drink the potions. Maybe the amount of potions should be roughly defined?

12Metatron:
what do you think is a good amount? number of players divided by 6 maybe?

McTsts:
I'm not sure

venomousbirds:
I don't think people will drink the potions. There's a small chance you get protection, but being unavailable for both polls could be seen as a loss. The chance to lose the chance elected position or get killed is too big to want to drink the potions. The night protection is often not neccesary for a lot of roles, and the risk is too big.

venomousbirds:
This is also what's wrong with chef, and people most of the time don't eat because there's an unneccesary chance at being killed/no risk in not eating

McTsts:
We've decided to not add this role. Chef already struggles to get their food eaten and they have reason to actually help certain teams.

PR #329, opened 2021-05-18
Create neighbor

McTsts:
Based on:
> What about giving citizen a “friend circle” (like list of names at start of the game) that includes another citizen. That way they could try to find other citizen, might make it more interesting

A TM role that shouldn't be confirmable but is more interesting than Citizen


Shepherdwhen:
I like it as it's a weaker town crier

McTsts:
It's similar to TC, but it's very easy to fake

McTsts:
This role will not be added in v13

cryptidjayden:
don't think it's really needed, to be honest. games with cits in tend to be very solvable anyway, no need to make it easier.

PR #330, opened 2021-05-18
Create amnesiac

McTsts:
based on:
> we can also consider a variant of the amnesiac thing some games have, but make it so you can only be town and learn the role relatively early. So basically you don't know your role, only that you're town, and then at the end of D1 you learn your real role or something. sound kinda dumb but it's an easy role to claim because it buys you time until you need to actually claim

and
> citizen, but changes role to another town role at the end of D1

another weak and claimable TM


Shepherdwhen:
very easy to fake claim, and makes the tm claims post D1 harder to trust
I like it

McTsts:

PR #331, opened 2021-05-18
Create Herald

McTsts:
based on:
> At the beginning of the game, choose a role. When there are no more players with that role left, then the Quaker is revealed to the public. If no one has the specified role, the Quaker will never be revealed.

a more claimable variant of the Quaker suggestion


Shepherdwhen:
once the quaker has been informed that the last player with the role they selected has died does it no longer get informed/can it no longer switch to another role?

I'm having a hard time understanding how this one works

McTsts:
> once the quaker has been informed that the last player with the role they selected has died does it no longer get informed/can it no longer switch to another role?
> 
> I'm having a hard time understanding how this one works

I don't know actually. Could both be one use or multiple uses. The original was one use, but this version could also be used multiple times.

voidmist:
I don't see why this _shouldn't_ be added lol. Its a pretty neat role added late game town strategies that can keep the wolves' with fakeclaims on their toes. But probably becomes undermined with the reporter.
The Quaker is likely useful early game too, to catch 1x role deaths or lone wolf shenanigans. I see it being really lost in closed rolelists though.

voidmist:
> > once the quaker has been informed that the last player with the role they selected has died does it no longer get informed/can it no longer switch to another role?
> > I'm having a hard time understanding how this one works
> 
> I don't know actually. Could both be one use or multiple uses. The original was one use, but this version could also be used multiple times.

Since the Quaker is getting revealed publicly, it sounds like a one time use. You can't really reveal the Quaker multiple times, can you.

McTsts:
> > > once the quaker has been informed that the last player with the role they selected has died does it no longer get informed/can it no longer switch to another role?
> > > I'm having a hard time understanding how this one works
> > 
> > 
> > I don't know actually. Could both be one use or multiple uses. The original was one use, but this version could also be used multiple times.
> 
> Since the Quaker is getting revealed publicly, it sounds like a one time use. You can't really reveal the Quaker multiple times, can you.

My suggested version doesn't have a reveal.

McTsts:
The suggestion is a modified version of the original version. I liked the basic idea, but we need more unconfirmable roles, so I modified it.

Shepherdwhen:
If it can be used multiple times it sort of acts as a weaker reporter, if it targets 1x roles in a list and then changes when a 1x role dies, also it can be confirmed by reporter that way

McTsts:

McTsts:
I have just realized this name makes no sense so I think this role should probably get a fitting name that actually makes sense

Shepherdwhen:
visionary?

McTsts:
I feel like that might sound more like the modern version and not the actual visions version, but I think that's probably a good direction

McTsts:
But we might've also already used up all similar words... :(

Shepherdwhen:
Prophet - Predicting death?
Dreamer - The death of a player comes in a dream?
Theorist - They theorise wo was the last of their target role?

McTsts:
Dreamer maybe

McTsts:
Name changed to Herald as suggested by @cryptidjayden

PR #340, opened 2021-05-19
Hags - Unique UA Suggestion

McTsts:
This is a somewhat complicated idea for a unique UA.
I was thinking we could have a UA that originally can win with everyone and just need to survive, but every time a team "wrongs" them, they turn against that team and can only win with the other teams. That'd basically be an UAA and while it can have late-game alignment change, this role would only lose allies and never turn enemies into allies, so I think this sort of late-game alignment change is fine. 
Basically several of these would exist at the same time, and if one is killed, the remaining ones turn against the team that killed them.

Why would anyone ever "wrong" them? - It's an easy safeclaim for town roles. (But wolves can also easily claim it)
But is it actually claimable? - They don't know each other, and the alignment changes are public, hence it's fairly easy to claim.
Would this be boring to play? - I think it could probably be interesting to play, but that is something that would require testing.

**Normal Hags:** Vicious Hag & Cautious Hag are the normal hags, but there could be 3-4 different ones in total if there's any interesting powers for them.

**Bitter Hag:** Bitter Hag exists because the hags would have no way of winning once all teams have wronged them. (Since UA aren't considered in the win conditions of other teams) Hence they change into a Solo team to still be able to win. Additionally, their spells are reset so that saving spells for the chance that they become solo isn't encouraged.


MrTurtleJr:
You accidentally wrote "The Cautious Hag may use their spell to attack another player." as opposed to protect under Details.

voidmist:
This is a really cool concept, I love it! This would work really well in larger games, and puts all factions on edge with choosing their targets. I like how town, generally having more power roles, actions per night, and lynch majority, are more likely to wrong the hags. This gives an implicit advantage to all the minority factions

In the case of a redirected kill (flute apprentice) or a wager kill, which party would wrong the hags? Also should public role reveals such as the ranger/jury be considered as wrongs?

McTsts:
> You accidentally wrote "The Cautious Hag may use their spell to attack another player." as opposed to protect under Details.

Oops

> This is a really cool concept, I love it! This would work really well in larger games, and puts all factions on edge with choosing their targets. I like how town, generally having more power roles, actions per night, and lynch majority, are more likely to wrong the hags. This gives an implicit advantage to all the minority factions
> 
> In the case of a redirected kill (flute apprentice) or a wager kill, which party would wrong the hags? Also should public role reveals such as the ranger/jury be considered as wrongs?

Public Reveal: revealing the hags doesn't wrong them, and I don't really see why it should.
Redirections: According to the redirection rules the person that initially sent out the action is still responsible fo the action (and not the redirector)
Wager Kill: The wager kill is executed by the devil.

McTsts:
Since Cupid loses their status as unaligned, maybe hags should also?

Shepherdwhen:
Would that make hags an UAAM (**U**n**a**ligned **A**lign **M**iscellaneous)?

McTsts:
what?

McTsts:
Hags are UAA and the only UAA other than Cupid to keep their role while changing alignment. Cupid is no longer considered Unaligned once they've selected a lover (they've aligned, hence the UAA), so I'd argue Hags should work the same way.

Shepherdwhen:
Oh I thought hags were UAM, due to them having the extra abilities

PR #341, opened 2021-05-19
More special lists

McTsts:


venomousbirds:
These are for the Role Book, right?

McTsts:
This is primarily a PR because when I merge it I'll need to set aliases and I can't while ingame

Shepherdwhen:
Should cult be added as one too?

McTsts:
> Should cult be added as one too?

Cult already has one (`$i cult`)

PR #342, opened 2021-05-20
Update vampire

McTsts:
Demonization disappears if no members of the underworld are left.


Mjnstag:
do we also need to add this to the undead description?

McTsts:
Demonization is something the Vampire does, so I don't think it needs to be part of the undead description

venomousbirds (approved):

venomousbirds (approved):

Mjnstag (approved):

PR #345, opened 2021-05-24
Create socialite

bobloy:
Socialite was inspired by a conversation on the discord.

It's a group role that joins all the available townsfolk groups. It can also be used in games without groups with it's secondary ability to act as any role for role guessing and journalist games.

### DIFF
--- a/categories/Townsfolk Group
+++ b/categories/Townsfolk Group
@@ -4,4 +4,4 @@ Butcher
 Cult Member
 
 *Juror*
- 
+*Socialite*
--- a/limited/socialite
+++ b/limited/socialite
@@ -1,4 +1,4 @@
-**Socialite** | Townsfolk Group
+**Socialite** | Townsfolk Group | Limited
 __Basics__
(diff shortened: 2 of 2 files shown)


McTsts:
I don't think this is a good idea. Usually, we only do one group at a time, so this would basically just be a member of that group with extra abilities, which I don't think is the point of this role. (Regardless, I also personally don't like the idea of a role being part of all those groups anyway)

https://github.com/McTsts/Werewolves-Roles/blob/6b704cfbf7b41489b31508d8309741d7d7f6522e/limited/socialite#L6-L7
These parts seem interesting, but I'm not sure if they really work as a role on their own. (Basically the idea that if a player says/guesses a role Socialite is a match)
However, I don't think this can be implemented in a reasonable way. Abilities like these have been reduced to weak disguises, strong disguises, redirections, and obstructions, and I'm not sure if anything like that can be used here. An obstruction would work on PW/CS, but not on Receptionist.

PR #350, opened 2021-05-25
Create Candlestick maker

Shepherdwhen:
An Idea for a Townsfolk group role that doesn't start as a group and instead makes it's own group

### DIFF
--- a/limited/Candlestick maker
+++ b/limited/Candlestick maker
@@ -1,4 +1,4 @@
-**Candlestick Maker** | Townsfolk Group
+**Candlestick Maker** | Townsfolk Group | Limited
 __Basics__
 The Candlestick Maker don’t start with more than one player in their group however each night they can add a player to their group
 __Details__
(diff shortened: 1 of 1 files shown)


venomousbirds:
This seems like Innkeeper but easier to get killed

Shepherdwhen:
It's a bit different as:

- It doesn't have anonymous chat
- It loses control of who is added if there are more non town than town
- It can remove one person rather than everyone at once
- It doesn't delete what was said

voidmist:
Consider naming the role Chandler:

> chandler (noun)
> _especially in the past_, a person who makes candles:
>
> [3rd definition](https://dictionary.cambridge.org/dictionary/english/chandler)

Shepherdwhen:
> Consider naming the role Chandler:
> 
> > chandler (noun)
> > _especially in the past_, a person who makes candles:
> > [3rd definition](https://dictionary.cambridge.org/dictionary/english/chandler)

I based the name on the fact we have a butcher and a baker already and an old rhyme, one in the UK especially, has a line that goes "The Butcher, The Baker and The Candlestick maker"

McTsts:
We've decided to not add this role. It seems fairly similar to Innkeeper and doesn't seem like it'd add anything unique.

PR #353, opened 2021-05-26
Create Firebug

Shepherdwhen:
The Firebug is currently a, mini wolves only, weaker version of a Pyromancer

This version updates the role the Firebug to be a role to help balance Pyros for games where 1 extra pyro would be too much and without the pyro they would be too weak

### DIFF
--- a/categories/Solo Power
+++ b/categories/Solo Power
@@ -7,3 +7,4 @@ Vampire (Underworld Team)
 *Ice Queen (Ice Royals Team)*
 *Reaper (Nightmare Team)*
 *Shepherd (Flock Team)*
+*Firebug (Pyro Team)*
--- a/categories/Solo Team Pyro
+++ b/categories/Solo Team Pyro
@@ -2,3 +2,4 @@
 Pyromancer
 
 *Wisp*
+*Firebug*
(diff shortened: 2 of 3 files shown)


McTsts:
The issue with this is that it's a lone solo, meant for bigger games, but usually we prefer having groups of 2 in bigger games.

Shepherdwhen:
the problem with Pyros is there are games where like 2 pyros is not enough but 3 pyros is too much and a wisp won't be balanced as it doesn't give pyros any more powders, this role is here to solve that dilemma by making a weaker powder role for pyros

McTsts:
It might solve that issue, but will it be fun to play as?

Shepherdwhen:
It has abilities tho help it find the pyros, it is also able to powder at night so has an active ability, it also can become a pyro and since it's a Solo killing is immune to an Aura teller, it will be a pretty fun role to play as while still being weak enough to be a balancing role

though it should probably be tested first

McTsts:
oh yeah it's immune to AT, interesting.

Shepherdwhen:
The role definitely needs testing but it could work as a limited role at least if not a full role

> oh yeah it's immune to AT, interesting.

and yes since it can't directly kill until it becomes a pyromancer it is a solo power so is not seen as a threat to AT

McTsts:
This role has been denied by GM vote. I think it's unlikely that we will consider a role like this necessary currently, so I don't think revamping this role is gonna increase it's chance, so I'll be closing this.

PR #364, opened 2021-06-02
define "success" for attacks

McTsts:


Shepherdwhen:
Probably put a thing that attacks can be successful even if not on the intended target, for fpa and hooker's target being attacked and either redirecting or surviving but hooker dying

McTsts:
> Probably put a thing that attacks can be successful even if not on the intended target, for fpa and hooker's target being attacked and either redirecting or surviving but hooker dying

Good idea! Added it.

Issue #372, opened 2021-06-04
Role Idea: Blackmailer


McTsts:
![image](https://user-images.githubusercontent.com/24660095/120818793-3b226780-c553-11eb-9dc0-b9811cd69315.png)

MrTurtleJr:
Interesting concept. A bit surprised that it is Unaligned, as it seems like a Wolf role to me. An Unaligned version would need some sort of goal? I feel like a Wolf (or Townsfolk, I guess) version has more purpose.

Townsfolk Blackmailer: Crucial when the numbers of Townsfolk are low; can prevent sniping (or unify Townsfolk) if used correctly; **but** only really works if there are TK's left to take out said Wolves, as else they'll just overpower the Townsfolk.
Wolf Blackmailer: Can make a Townsfolk seem suspicious by voting incorrectly; learns new role; **but** could have a "clearing" effect, where this person now is trusted and the Wolf gains next to nothing (though it could work like Sab. Wolf where Wolves are frequently self-targeting and maybe even offers an excuse to setup a snipe).

Needs some more working on, as I can see flaws with both alignments.

McTsts:
I think the Wolf one may have more potential. Town already has Stalker and Grandma, and while this one is a unique variant I think it would be better for the wolves who don't have a role like this. Also, we don't really need more confirmable town roles currently.

MrTurtleJr:
Yeah, I definitely agree. I'm leaning Wolves, as it seems to better fit their "agenda".

12Metatron:
it works as a wolf. but the problem with anonymous discussion stuff is that it becomes very easy to narrow down who it might be, which might make it a role that backfires often. But I do like all the strategies it opens when it comes to making deals with solo, due to being able to expose their role. If that part becomes a part of the final role and its a wolf role, I think that they should be able to choose to announce it publicly after they see it for themself first

McTsts:
> it works as a wolf. but the problem with anonymous discussion stuff is that it becomes very easy to narrow down who it might be, which might make it a role that backfires often. 

They could always also just choose not to talk at all if they want to. But that actually made me realize that wolves don't have any whispers role at all so having a wolf whisper role would actually be pretty cool.

> But I do like all the strategies it opens when it comes to making deals with solo, due to being able to expose their role. If that part becomes a part of the final role and its a wolf role, I think that they should be able to choose to announce it publicly after they see it for themself first

I'm not sure what you mean by that.

[...]

PR #384, opened 2021-06-05
Blackmailer Suggestion

McTsts:
This is based on Bart's Blackmailer from #372 

Instead of UA/Town this one is Wolf as it seemed to make more sense. Also, the wolves currently don't have a whispers role.
As this role is a non-wolf (like e.g. Tanner) I made it a packless one (which also fits with the reveal ability it has), but that could be changed (though it should probably have a different name then).

### DIFF
--- a/werewolves/power/blackmailer
+++ b/werewolves/power/blackmailer
@@ -1,0 +1,9 @@
+**Blackmailer** | Werewolf Power
+__Basics__
+Each night, the Blackmailer may blackmail another player into voting a certain way.
+__Details__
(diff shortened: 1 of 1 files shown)


McTsts:
Hmm, actually if this is a non-investigative role it shouldn't be affected by weak disguises.

McTsts:
image

McTsts:
This is apparently not particularly useful, so it would need to be changed.

McTsts:
This idea didn't work out.

PR #386, opened 2021-06-05
Make WWW imitate most common WM

McTsts:
This would be a small but nice buff for WWW which doesn't affect the role lists as we currently design them and simply adds more possibilities. (So because usually in WWW the most common WM is Wolf the WWW would imitate Wolf with this which is exactly the same behavior it has right now)

With this buff the following variants would be possible:
White Hellhound
White Scared Werewolf (we already used this one once)
White Sneaking Wolf
White Wolf (White Werewolf)
White Wolf's Cub (we already had this one once)

technically also White Bloody Butcher & White Knowledge Wolf but those are unlikely.


venomousbirds:
I actually think this is a very interesting idea

PR #387, opened 2021-06-06
Rename Infected Wolf to Infecting Wolf

### LINKED ISSUE #367: Infected Wolf Rename
image

It was previously suggested to rename Infected Wolf, but renaming it too much would probably be confusing. This suggestion is not too confusing but probably makes more sense than the current name.

McTsts:
> It was previously suggested to rename Infected Wolf, but renaming it too much would probably be confusing. This suggestion is not too confusing but probably makes more sense than the current name.

Fixes #367


bobloy:
Infective wolf? More like _ineffective_ wolf, amiright? 

Also, maybe Infecting Wolf?

McTsts:
Infecting Wolf also works

PR #404, opened 2021-06-19
Make Guardian protect from NKs

### LINKED ISSUE #397: Guardian makes no sense
Guardian is the weakest / least useful of the elected role, but it's also the last one meaning it's entirely worthless - it's not even relevant to get the next elected role (like mayor sometimes is) either another elected role should be added or guardian should be buffed, probably.

McTsts:
Fixes #397

### DIFF
--- a/elected/guardian
+++ b/elected/guardian
@@ -2,7 +2,7 @@
 __Basics__
 The Guardian protects the town. When there are 15 or fewer players, the Guardian may choose one player. This player cannot be killed by attacks during the current night. When there are more than 15 they may choose up to two.
 __Details__
-The Guardian chooses their protected player(s) during the night, and they will be protected from attacks durring the current night. 
+The Guardian chooses their protected player(s) during the night, and they will be protected from attacks during the current night. 
 They may not protect the same player two days in a row, but may protect themself and/or alternate between two or more players. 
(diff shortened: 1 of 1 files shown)


McTsts:
Possibly announce who's protected?

Shepherdwhen:
> Possibly announce who's protected?

Announce who was protected the night before?

McTsts:
No, like the current guardian announces who's protected so lynching them isn't even an option (and you can't fail to lynch them) so it'd make sense to keep that part. Also, it's probably more interesting if you actually have to reveal who you're protecting.

McTsts:
And, of course, it's not quite as powerful.

Shepherdwhen:
I feel like part of it is wasting the enemy kills, if you announce who is protected before the kill happens the killer will kill someone else

Shepherdwhen:
also this would just out the powerful roles to wolves

Shepherdwhen:
meaning you couldn't protect a TI or anything only the mayor or reporter or guardian

McTsts:
> I feel like part of it is wasting the enemy kills, if you announce who is protected before the kill happens the killer will kill someone else

But that's not part of how Guardian works. Current Guardian isn't meant to waste votes.

PR #407, opened 2021-06-21
Create Poisoner

bobloy:
A werewolf role that can beat protection roles, but choreographs the target

### DIFF
--- a/categories/Werewolf Power
+++ b/categories/Werewolf Power
@@ -8,3 +8,4 @@ Tanner
 
 *Corrupt Wolf*
 *Pack Tanner*
+*Poisoner*
--- a/limited/poisoner
+++ b/limited/poisoner
@@ -3,5 +3,5 @@ __Basics__
 Each night, the Poisoner may poison a player, weakening them.
 __Details__
 Each night, the Poisoner may poison a player. Poisoning a player is an immediate ability. The player is informed they are poisoned.
-The following night, the poison will disable passive protection abilities. This makes roles like Macho susceptible to night kills.
+The following night, the poison will disable passive protection abilities. This makes roles like macho susceptible to night kills.
 The Poisoner is not a lycan, nor are they a member of the wolfpack.
--- a/limited/poisoner
+++ b/limited/poisoner
@@ -1,0 +1,8 @@
+**Poisoner** | Werewolf Power | Limited
+__Basics__
+Each night, the Poisoner may poison a player, weakening them.
+__Details__
+Each night, the Poisoner may poison a player. Poisoning a player is an immediate ability. The player is informed they are poisoned.
+The following night, the poisoned player may not take an action.
+Additionally, poisoning will disable passive protection abilities. This makes roles like Macho susceptible to night kills.
+The Poisoner is not a lycan, nor are they a member of the wolfpack.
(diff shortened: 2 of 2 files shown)


Shepherdwhen:
Roleblocking roles are a sin in wwr

just saying

bobloy:
Removing the "can't take an action" part still leaves a perfectly acceptable role IMO, if necessary.

McTsts:
I thinking disabling a passive protective ability is acceptable. I'd be against disabling active protections or blocking active powers. I think this would work as a role, though it would probably be a rather rare one.

Shepherdwhen:
would this include devil/demon protection? that passively activates on attack

bobloy:
It would be easy enough to say it works on town roles only to avoid devil/demon stuff.

bobloy:
Further, if its only town roles, I could even list all the roles it works on.

Runner
Acrobat
Macho
Cursed Civilian

McTsts:
I like to avoid lists of roles because they're not very future proof and we have a high chance of forgetting to update them. E.g. if/once we add Idiot it would need to be added to this list. I think just saying it only affects town roles (or maybe town and wolf roles?) would be best. Maybe it should also have some mechanic like Tanner that gives it a bit of a chance to find the wolfpack

> If the tanner disguises a player who is part of the wolfpack, the tanner is informed by the Hosts that “At least one of the players you tanned today is part of the wolfpack”.

Shepherdwhen:
idiot too

bobloy:
> I like to avoid lists of roles because they're not very future proof and we have a high chance of forgetting to update them. E.g. if/once we add Idiot it would need to be added to this list. I think just saying it only affects town roles (or maybe town and wolf roles?) would be best. Maybe it should also have some mechanic like Tanner that gives it a bit of a chance to find the wolfpack

Either town/wolf roles only, or because this wording is used:
`If the Devil is attacked while a soul is in the pool, a soul will automatically be used to protect the Devil. `
It implies it not passive, but a triggered active ability.

McTsts:
It's probably best to just say it only affects town/werewolf roles

McTsts:
Maybe this should be a day ability that affects the next night instead? Seems like that'd make more sense

bobloy:
> Maybe this should be a day ability that affects the next night instead? Seems like that'd make more sense

That would buff them, allowing a night 1 poison. This prevents anyone from being weakened until night 2

McTsts:
This role received controversial responses due to it being sort of like role blocking. No conclusion has been reached for v13.

venomousbirds:
I dislike the role blocking part, and completely disabling those abilities just takes away from those roles. People would not like playing a game with this role, I think

bobloy:
I think this roles adds a lot more interesting gameplay. If I were macho and I were poisoned, I'd have to decide whether to reveal that I was poisoned or not, betting whether the poisoner has found the wolves.

This could also lead to wolves taking a risk and attacking a macho mayor. 

Also, I like to think of roles like these from the perspective of the games they're not in. This role existing allows fake claiming macho to be safer for wolves.

McTsts:
Closing this suggestion due to lack of development on it. Feel free to reopen this (now or at any point in the future) if you want to further discuss/develop this idea.

PR #410, opened 2021-06-25
Create Matchmaker

bobloy:
Matchmaker is a solo killing role that doesn't have a team.
Each night they make matches out of different players, causing them to die when their match dies.

The matchmaker may only kill 3 times per game, forcing them to strategically make matches for most of the game.
(Number of kills maybe should scale with game size)

They are night kill immune to help them survive without a team, and win the 1v1 scenario with a wolf.


McTsts:
Wouldn't they just match players and then attack one of those matched players?

McTsts:
Oh wait, I think I get it

McTsts:
We think that this would probably end up just being some sort of alternative pyro where they chain together their matches and then use their attack to kill an entire chain.

venomousbirds:
I don't like it, it's another solo killing that just takes a bit from other roles. Cupid and Pyro, it seems. The night immunity is completely unneccesary, and I don't think people would like playing with a role like this ingame

bobloy:
I don't think this behaves nearly as much like pyro as you would think. Unlike pyro, it can't be kept hidden until the end of the game. If the wolves hit someone you've matched, it reveals info that a matchmaker may be in the game.

Also, Cupid is a much different role with wildly different outcomes from a single choice. Yes they both "match", but they're very different effects.

Yes chaining could be the primary use, but making more matches outside of a chain is the fastest way to kill people, so the person playing this would have to balance that. 

I can't imagine people would dislike playing against this role anymore than playing against pyros. And the night immunity balances that this is one of the only role in the game without a team at any point.

bobloy:
I think this role would be best as the only solo, with roles like bard, cult, assistant, and hooker.

McTsts:
Closing this suggestion due to lack of development on it. Feel free to reopen this (now or at any point in the future) if you want to further discuss/develop this idea.

PR #419, opened 2021-06-28
Amnesiac Changes

venomousbirds:
I didn't think Amnesiac fit too well into the ww theme, so I propose we change it to Nitwit. I also thought the description was a bit messy and hard to read, so I edited it slightly.


McTsts:
The description is a lot better this way. I still prefer Amnesiac but Nitwit is acceptable if everyone else prefers it this way.

PR #420, opened 2021-06-28
Update thief

Shepherdwhen:
Just puts the win condition of thief on one line rather than having it all over the place on 2 separate lines at the top and bottom of details


venomousbirds:
Put the alternative wincon (no players left to switch to) as the second win con. Surviving is the main wincon

Shepherdwhen:
> Put the alternative wincon (no players left to switch to) as the second win con. Surviving is the main wincon

done

McTsts (approved):

venomousbirds:
small change to make the wincon a bit more clear

McTsts (approved):

PR #434, opened 2021-07-24
CH protect against any wronging

### LINKED ISSUE #430: Cautious Hag Spell
maybe Cautious could protect against any type of wronging

McTsts:
Fixes #430


McTsts:
resolved conflict

venomousbirds:
I assume this is only specific to CH protecting another Hag?

McTsts:
I believe it was meant that any player could be protected from any effect that would count as wronging if it was used on a hag

McTsts:
Though that specific power could also only be specific to hags (that was the purpose of the change)

venomousbirds:
I think it would make more sense for it to only count for hags, I don't think it makes sense to make a protect that is as strong as this for just anyone

McTsts:
Changed it

venomousbirds (approved):
Sounds good, hags being able to help eachother like this and staying unwronged for longer is a nice addition

PR #436, opened 2021-07-24
IW edge case list

McTsts:
Fixes #428

PR #437, opened 2021-07-24
IW failure on hit packless werewolf

### LINKED ISSUE #427: IW => Packless Werewolf
suggested special case for IW if hit packless wolf: nothing happens (useful for IW=>Imp)

McTsts:
Fixes #427

This is useful in general but particularly for IW => Imp

### DIFF
--- a/werewolves/power/infecting wolf
+++ b/werewolves/power/infecting wolf
@@ -3,5 +3,5 @@ __Basics__
 During Night 1, the Infecting Wolf must choose one player to convert to a wolf.
 __Details__
 If the Infecting Wolf fails to choose a player during Night 1, a random player will be selected for them. This ability can only be used once. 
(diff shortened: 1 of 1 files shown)


Shepherdwhen:
would #436  have this as part of it too?

venomousbirds (commented):
Would make Infecting Wolf even more careful, not sure if that's what we wanna go for but I don't dislike the change

McTsts:
> Would make Infecting Wolf even more careful, not sure if that's what we wanna go for but I don't dislike the change
No, I'd consider this a buff. Turning a non-pack wolf into a wolf is not an advantage. With this the Infecting Wolf will know they have hit the non-pack wolf and will be able to contact them.

venomousbirds (approved):
Didn't think about it like that

PR #450, opened 2022-04-30
Angel cancels following lynch

venomousbirds:
Similar to Despot cancelling the following election, Angel will cancel the following lynch whenever they are lynched. This might make Angel harder to play
Will need reviewing after OS7C

### DIFF
--- a/unaligned/miscellaneous/angel
+++ b/unaligned/miscellaneous/angel
@@ -2,7 +2,7 @@
 __Basics__
 The Angel wins when they are lynched.
 __Details__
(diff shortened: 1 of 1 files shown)


McTsts (approved):
Seems like a good idea. I don't have the data on Angel winning too much, but it seems believable, and this might help. It makes the role more advantageous to the wolves, but Angel is already wolf-sided so this is probably fine.

MoltenAshes:
I like the change. Although this role will be more wolf-sided than town-sided now, I personally believe a slight buff to the wolves whenever this is in rotation would be a nice change in pace.

PR #451, opened 2022-05-02
Rename "Uncertain Hag" to just "Hag"

McTsts:
I think in any scenario where you use hags in the role list you should instead use this role (else the claim space is not as great), Uncertain Hag is a stupid name and was meant to be temporary. I think "Hag" sums the role up well enough. It's a Hag without a specialization (so far) and it doesn't sound kinda stupid like "Uncertain Hag"


venomousbirds (approved):
Don't see a problem with this

PR #453, opened 2022-05-02
An improvement of Plague Doctor

### LINKED ISSUE #449: Plague Doctor
There is nothing the Plague team can do against the Plague Doctor, and since Plague team is the PD's only enemy nothing is stopping the PD to just say "I'm the PD" at the start of the game and be fine for the entire game. To be fair any other person could also claim to be the PD, but I'm not sure if that would actually happen. The PD can basically instantly public claim and it's probably somewhat difficult for anyone to cc the claim.

McTsts:
Fixes #449 

There is nothing the Plague team can do against the Plague Doctor, and since Plague team is the PD's only enemy nothing is stopping the PD to just say "I'm the PD" at the start of the game and be fine for the entire game. To be fair any other person could also claim to be the PD, but I'm not sure if that would actually happen. The PD can basically instantly public claim and it's probably somewhat difficult for anyone to cc the claim.

This change makes it so the plague team can actually do something against the Plague Doctor, meaning the Plague Doctor has to actually *play* instead of announcing their role and just surviving (this is in fact what happened when we previously had the old version of it in the game). The Plague team *cannot* be punished for targetting the Plague Doctor as then they'd have no reason to do so, and the PD could once again just announce their role. If the Plague team can kill the PD but needs to use up an rodent while doing so, the Plague team doesn't really profit from it. People targetted by rodents should *survive* as long as possible. So the idea is that they don't waste a rodent, merely a night (wasting a single night, but not a rodent is absolutely worth it).

I also propose to remove the voting thing. This is an leftover from when the Plague Doctor was the Immune Survivor (that being a variant of Survivor), and is no longer necessary with my proposed changes.

With my proposed changed the Plague Doctor can actually participate in the game in a meaningful, interesting and unique way. Already the previous changes to Plague Doctor give it a pseudo-investigative power, and with my proposed change this can be used to its fullest.
The Plague Doctor would think of the most likely targets for the plague, then check them. If they *have* the plague they can't be part of the plague team. This means the plague doctor could form some sort of alliance of players from different teams against the Plague team, which could potentially be an interesting mechanic. Furthermore, now that the PD won't immediately publicly claim on Day 0 this makes this role quit a bit more claimable.  

This still needs testing, but I can't imagine it could be less interesting than that one game with the old version of the role we played on vip.

### DIFF
--- a/limited/plague bearer
+++ b/limited/plague bearer
@@ -11,3 +11,4 @@ Any player who visits or is visited by a contagious player is also immediately c
 At the start of the second night after being contaminated the player is informed that they are contaminated.
 At the end of the second night after being contaminated the player will die (cannot be avoided, not counted as an attack).
 If no members of the plague team remain all players are no longer contaminated, however they are not informed about this.
+If the Plague Bearer attempts to use a rodent on the Plague Doctor, the attempt fails, the Plague Doctor dies (unavoidable) and the rodent is *not* used up.
--- a/limited/plague doctor
+++ b/limited/plague doctor
@@ -4,5 +4,5 @@ The Plague Doctor is mostly immune to the plague bearer’s disease and requires
 __Details__
 As soon as the plague team is eliminated, the Plague Doctor ascends and wins.
 Each night, the Plague Doctor may check a player to see if they have the plague and, if they do, since when they have it. This is an immediate ability.
-If the plague bearer attempts to contaminate the Plague Doctor with one of their rodents, the Plague Doctor immediately dies.
+If the plague bearer attempts to contaminate the Plague Doctor with one of their rodents, the Plague Doctor dies (unavoidable).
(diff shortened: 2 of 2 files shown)


McTsts:
One potential flaw in this suggestion is that the PD's targets will actually betray the PD, as they'd prefer the Plague team to target the PD instead of themselves. I'm not sure if I can really determine if this is likely to happen without seeing it in action. A potential fix for this would be to give the Plague Doctor a certain amount of healing powers (e.g. be able to cure a player once per 12 players in the game) (12 being twice as much as the 6 of the plague team's rodents)

McTsts:
Added a curing power

McTsts:
There have been concerns about the PD's instant death feature causing inactivity for the PD (they'd hope to have higher chances by doing nothing, because doing so would decrease their chances of being revealed). I don't think this will happen, but I want to note it for future reference.

PR #454, opened 2022-05-02
Remove Tasty Villager

McTsts:
This doesn't seem to be a very popular role. The role's main argument for existing is apparently "it's good for balancing", but honestly it seems like it'd be terrible to play with, so I think we're better off not having it.


venomousbirds:
i didnt like this role to begin with so i definitely dont mind removing it

Issue #251, opened 2021-05-05
Suggestion: Trickster x Infected Wolf


McTsts:
An infected trickster turns into a trickster wolf instead

Mjnstag:
would trickster wolf have negative votes on both public/private polls?

McTsts:
> would trickster wolf have negative votes on both public/private polls?

Probably

Mjnstag:
So that would mean that if the trickster wolf cannot kill anymore using the wolfpack poll if they are the last one alive?

Shepherdwhen:
Why would it ever vote on wolfpack poll except in WWW situations where it is wolf + TW + WWW and pack want to kill WWW

Shepherdwhen:
If it has -1 private

McTsts:
> So that would mean that if the trickster wolf cannot kill anymore using the wolfpack poll if they are the last one alive?

Seems fair for balancing.

McTsts:
> Why would it ever vote on wolfpack poll except in WWW situations where it is wolf + TW + WWW and pack want to kill WWW

last minute changes, same as current AW really

Mjnstag:
thats a good point for when members are asleep during phase change

Shepherdwhen:
I think it works, do we do this with executioner and stalker too?

McTsts:
No, I just suggested it as I thought it'd fit the "Trickster" role. Also both Stalker/Executioner Wolf would be a definite advantage. Trickster Wolf is not necessarily.

PR #456, opened 2022-05-02
Trickster Wolf

McTsts:
Implements #251


venomousbirds:
interesting way to incorporate a role thats otherwise really stupid (trickster wolf) into something that makes more sense

PR #461, opened 2022-05-03
Add Knowledge Wolf

McTsts:
Add Knowledge Wolf as a full role. We've used it quite a few times, so I think it only makes sense to add it to the proper roles. It's also not really something we're gonna change.

Limited roles should be left for roles that generally aren't used without warning or roles that are still under development.

This role only makes sense in games with a partial role list, and it's better to not have limited roles in those imo (unless they are specifically named)


venomousbirds:
I don't really like Knowledge Wolf, it feels incredibly out-of-character for werewolves, especially considering it was only made as a joke for mini wolves

McTsts:
It's been used a few times (S4G5, S6G5 (A&B), S7G2, V2G2, V2G5 (twice in this game, supposedly?)). I think the general use of the role isn't to tell the wolves the full role list (which is what the Mini Wolves concept for the role was) and instead to just give them a small amount of extra information. Maybe the role could be changed to reflect this?

venomousbirds:
A way to make it so Knowledge Wolf doesn't get the role list but still gains info nobody else can while also staying within the vibe of Werewolves sounds hard and I honestly can't think of anything right now. 
I just don't like how they just get a list. Maybe some information every night instead of a list d0 would make more sense?

McTsts:
I meant mainly to make it so that "knowing the full role list" isn't part of it anymore (that was the original Mini Wolves role), instead it would only get a few pieces of information at the start (I believe that is also how we've used it until now)

venomousbirds:
That's mainly how we used it but I still don't think it fits the Werewolves vibe.
Needs some others to review this

cryptidjayden:
I kind of like KW - complements the informed majority role of werewolves. It's been used to great effect formerly. The only thing I'd suggest is a potential name change?

Shepherdwhen:
Not sure if it makes sense, but what if GWW was a role like WWW that wins solo but if it's left alive with WWW in the final two it steals WWW's win. Kind of a double betrayal, not only do wolves lose but the WWW loses their win.

My thought process is the kill priority being GWW comes before WWW so GWW could win if in final 2 with WWW

McTsts:
> Not sure if it makes sense, but what if GWW was a role like WWW that wins solo but if it's left alive with WWW in the final two it steals WWW's win. Kind of a double betrayal, not only do wolves lose but the WWW loses their win.
> 
> My thought process is the kill priority being GWW comes before WWW so GWW could win if in final 2 with WWW

I think you're in the wrong PR.

McTsts:
> I kind of like KW - complements the informed majority role of werewolves. It's been used to great effect formerly. The only thing I'd suggest is a potential name change?

Yeah I feel the same way, which is why I wanted to make it a proper role. A name change is completely reasonable, this one has never been changed from its joke role origins

Shepherdwhen:
> > Not sure if it makes sense, but what if GWW was a role like WWW that wins solo but if it's left alive with WWW in the final two it steals WWW's win. Kind of a double betrayal, not only do wolves lose but the WWW loses their win.
> > My thought process is the kill priority being GWW comes before WWW so GWW could win if in final 2 with WWW
> 
> I think you're in the wrong PR.

oops lol I had like 3 open at once

cryptidjayden:
i think we could also have a packless version of kw (archivist and pack archivist?). idk just seems logical to me

McTsts:
Packless wolves without an ability to find or join the pack are kinda bad, so I don't think we should add one. That was the original reason Tanner got its invest ability.

McTsts:
Maybe to fit with the general way the foxes work, the fox learns a role (or several) every day and the pack learns the alignment?

McTsts:
Also going back to KW, it was used in S7G8 and since KW only makes sense in partial/private lists which optimally *shouldn't* use limited roles, I definitely think we should at least make KW a proper role. I'm not sure if I find "Archivist Wolf" to be a better name than "Knowledge Wolf". imo knowledge wolf seems more passive whereas archivist seems more active.

McTsts:
PR is now ready. Archivist Fox is now #527 instead. I think using "Archivist" for the fox would make more sense since that's actually an active power, whereas KW is a passive one.

venomousbirds:
Potential name changes:
• Wise Wolf (would be another WW for short though)
• Intelligence Wolf (cause it gains intelligence)
• Observing Wolf (Because it has 'observed' the role list)

McTsts:
I don't think Observing makes much sense. Intelligence and Wise I think make sense. Intelligence imo is far worse than Knowledge, it really doesn't sound like a serious role name that way. Wise Wolf I also considered, and I think it sounds alright and makes sense, but WW is not a great abbreviation.

cryptidjayden:
none of those really do it for me. maybe sage wolf?

cryptidjayden:
though that's SW which is also common

venomousbirds:
I think none of those suggestions you just made make sense, the last two sound very un-ww like, sage wolf might work but like you said, yet another SW

venomousbirds:
idk what augur is but if you add a k at the end (augurk) it means pickle in dutch

McTsts:
Yeah uh, haruspex or augur... no. I think Wise is probably better than Sage especially bc both abbreviate badly anyway. Still prefer Knowledge over all of them.

cryptidjayden:
yeah, i prefer knowledge over wise a lot. wise just sounds really wrong to me

venomousbirds:
I like Wise Wolf but the WW abbreviation is inconvenient. I just don't like KW as a whole so I'm biased against it lmao

cryptidjayden:
eyes wolf. bc it has eyes.

(this is a joke)

venomousbirds:
written like '👀 Wolf'

Shepherdwhen:
Does Informed wolf work?

venomousbirds:
It's definitely a possibility, I personally prefer Wise Wolf

McTsts:
I think it makes sense but I don't like it much. Of the ones I think make sense I'd say:

Knowledge > Wise > Informed > Intelligence

Shepherdwhen:
Why did we need a new name again?

venomousbirds:
because knowledge wolf kinda sucks in terms of names, i was only meant as a joke role in minis anyways
i still dont like it

ThomasRubycrust:
I'm unsure if we can find a better name than KW? And also thats actually a decently well known role name by now

ThomasRubycrust:
or like... Archive Wolf (i dont like it much but it would make sense if Archivist Fox exists, since Archive feels more passive I think)

ThomasRubycrust:
Potential Names:

Knowledge Wolf - Current
Knowledgable Wolf - A more proper name, but keeps it similiar
Knowing Wolf - A more proper name, but keeps it similiar; Worse than Knowledgable imo
Wise Wolf - Acceptable, but I hate WW as a abbreviation
Sage Wolf - Kinda bad imo
Archivist Wolf - Consistent with the potential Archivist Fox (though other foxes have no wolf equivalent)
Archive Wolf - Worse than Archivist imo, but fits better
Intelligence Wolf - The same 'bad' naming style as Knowlegde Wolf, but sounds worse imo. Worst option imo.
Intelligent Wolf - Same as above, but as a proper name. Acceptable.
Informed Wolf - Similiar to the one above, acceptable.
Perceptive Wolf - Doesn't fit as well as some others I think, but is alright
Observing - Same explanation as Perceptive, but I think Perceptive is better
Observer Wolf - No.
Perception Wolf - Same name type as Knowledge Wolf honestly
Insightful Wolf - Same explanation as Perceptive/Observing, but the connotation just seems wrong
Clever Wolf - Doesn't really fit all that well imo
Eyes Wolf - No
Haruspex Wolf - What? No
Augur Wolf - One of the worst ones imo
Scholary Wolf - Makes sense, but no

venomousbirds:
archivist is pretty okay, its not like all foxes need a wolf equivalent and what even is archivist fox i cant find it anymore
if its potential, i dont mind scrapping the name to rename kw

ThomasRubycrust:
> archivist is pretty okay, its not like all foxes need a wolf equivalent and what even is archivist fox i cant find it anymore if its potential, i dont mind scrapping the name to rename kw

Archivist Fox was a variant of KW discussed in this PR, its also now in issue #527

5uso:
"Insider Wolf", gets arrested for trading stocks

venomousbirds:
it will get confusing is the problem

if it does, i prefer fox over kw because i dont like kw. fox would be a great alternative

ThomasRubycrust:
> "Insider Wolf", gets arrested for trading stocks

We haven't added stock trading yet, but maybe in the future

Shepherdwhen:
> > "Insider Wolf", gets arrested for trading stocks
> 
> We haven't added stock trading yet, but maybe in the future

When we become a PLC we will be on the Stock Market





> it will get confusing is the problem
> 
> if it does, i prefer fox over kw because i dont like kw. fox would be a great alternative

I like the fox my only issue with it is it will likely be used less because the information will have to be pre planned, might not be useful when the fox gets it, i.e could get the info there is a Huntress in the game but the Huntress is already dead and revealed, and has to be balanced around how long the AF stays in the game as an AF, like and AF that dies N1 is much more useless than a KW but an AF who lasts to D5 could end up being stronger than a KW

ThomasRubycrust:
Yeah, Archivist Fox will need to be tested, but while it's similiar to KW it definitely isn't a replacement for it.

venomousbirds:
id love for it to replace kw i dont like kw

ThomasRubycrust:
I think we can discuss how AF would work in #527

I've now renamed KW to Knowledgable Wolf since it maintains the general naming and abbreviation, but is a more consistent name. Same as the Infected Wolf rename this should be relatively not confusing.

venomousbirds:
yup, same acronym is fine

venomousbirds:
would be great if we delete the role but sadly enough people are against it

venomousbirds:
thats literally what you can already do with kw

Shepherdwhen:
I'm just saying the way it's worded doesn't make it seem that way

ThomasRubycrust:
Ive previously suggested in this PR (iirc) that we change it to be that by default, instead of that being the alternative

ThomasRubycrust:
Because I think partial lists are the most common use of KW

venomousbirds:
sounds good to me

ThomasRubycrust:
> kw makes no sense in open games?

I meant that afaik the KW list is currently usually also partial

venomousbirds:
oh like that, ye
its better to have the partial be the default anyways

ThomasRubycrust:
Changed it to have a partial list as default

venomousbirds (approved):
I highly dislike this but most people seem to be in favor of keeping it and the role itself has been used more than other limited roles. I hate this but I guess I'll have to accept it 😢

PR #464, opened 2022-05-03
Remove acrobat

McTsts:
In our survey it was one of the least popular roles and it's basically never used. Doesn't really need to stay in the role book imo.


venomousbirds:
was meant to be a joke role anyways, dont think anyone will care

venomousbirds (approved):

PR #466, opened 2022-05-03
Remove Gray Werewolf

McTsts:
This is a very unpopular role that is rarely used. I feel like this role encourages some sort of gamethrowing? (The other play wins with the GWW but their team loses?) That's just a strange and bad mechanic.

### DIFF
--- a/archive/gray werewolf
+++ b/archive/gray werewolf
@@ -1,4 +1,4 @@
-**Gray Werewolf** | Solo Killing
+**Gray Werewolf** | Solo Killing | Archived
 __Basics__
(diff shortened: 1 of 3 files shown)


venomousbirds:
I think we can maybe change GWW, it would be nice to have another www-like role, but i do agree the wincon seems weird

McTsts:
It could be changed, but I'm not sure how? Right now it's just off-brand white werewolf

cryptidjayden:
Yeah, right now GWW is fucking useless outside of mini wolves. We can always create another WWW-like role later and name it that, but there's no point keeping this around til then.

McTsts:
Yeah, I agree. @venomousbirds is it fine if we remove it for now?

venomousbirds:
sure

PR #469, opened 2022-05-03
Remove Apprentice

McTsts:
This role has *never* been used. I don't think we are very likely to use it, so there's no reason it should remain in the role book.


McTsts:
@venomousbirds wants to try this role in a game at some point

McTsts:
We still have not used this.

Shepherdwhen:
> We still have not used this.

S7G6 used it idk how effective it was but it was still used

McTsts:
Oh huh, that was before I opened this the first time. Weird that I didnt find that usage of the role.

PR #470, opened 2022-05-03
What is alignment

### LINKED ISSUE #366: What is "alignment"?

McTsts:
Fixes #366 

Changes texts related to alignment in order to have on concept of alignment that is shared by all texts.


_categories/alignments
```diff
@@ -1 +1,2 @@
 Alignments are a big part of Werewolves Revamped. They decide your win condition (the way for you to win), and they decide who you can or can't cooperate with.
+Each role has its base alignment listed next to the role name and generally this is the same as the player's actual alignment, however rarely it is possible to change alignment without changing role.
```
venomousbirds, line 2: What's meant with "changing alignment without changing role" i cant think of a role that stays the same but changes alignment, they always have different names (cupids too)

McTsts: the cupid's lover changes alignment but not role

McTsts: arguably the hags also do, but its more of a "the alignment changes what it is" than a "the role changes what alignment it has"

venomousbirds: but it changes into a different cupid role, isn't that technically a role change?

McTsts: no, the cupid roles are only for the cupid. The lover doesn't change role.

venomousbirds: right i misread, but if that's the only exception, shouldn't it be specified?

McTsts: I don't like listing things because then in the future if we ever add anything similiar it's very likely we'll forget to update it here. However, I guess for this we could do it, because its somewhat unlikely we'll add something similar.

venomousbirds: anything remotely similar would probably be too similar to cupid to accept, so writing it down specifically doesn't have people wondering which roles exactly can change alignment without changing role

venomousbirds:
Reads a lot clearer, do I need to change the how-to-play messages?

McTsts:
Yeah, I think it should be updated in how-to-play as well.

Issue #324, opened 2021-05-17
Suggestion: New Role Idea [Golem]


12Metatron:
How it works is basically just despot but for guardian.

Despot is not a very fun role to play because they win too fast and are knocked out of the game very early. And if they miss their first chance then they have to wait a long time for another one, during which there's not much for them to do. In this role they would have more things to do, more ways to manipulate the game and be involved, and get to stay in the game longer. #

Ideas for consequence: for it to be a worthwhile consequence late into the game, i guess it could be that it uses guardian powers on a random voter or a chosen voter like angel (could be either, flexible on that part)

Shepherdwhen:
issue is there have been games which have never even reached the guardian stage as wolves never want reporter unless they are protected or a wolf then they never survive long enough to reach guardian election

12Metatron:
yeah so thats where the strategy comes in, they have to be involved in the game in order to achieve it

McTsts:
Sorry, I don't think is something we want to add.

ThomasRubycrust:
Hm, actually, couldn't this be interesting if buffed? Give the Golem the power to protect one player each night, which makes it so it has a real chance of getting the game to reach the Guardian stage and it would also then be a more town-sided role (unless the wolves immediately get Mayor D1, but that's fair imo). It could be good for games that don't have roles like Runner/Macho/Hooker.

Something additional to balance it could be that it learns if their defense was used and if so, they can't defend in the next night.

ThomasRubycrust:
@12Metatron I created a role prototype based on your idea.

PR #474, opened 2022-05-04
Create golem

ThomasRubycrust:
Implements #324


venomousbirds:
I haven't even seen this role before, but I really like the idea of an ua for every election. Problem is that Guardian doesn't happen much so it's another role that is best put in larger games
The nightly ability definitely does help to make it more interesting

ThomasRubycrust:
The rarity of Guardian is why the role suggestion was initially closed. With the nightly ability it should increase the chance of Guardian being able to occur significantly.

12Metatron:
i like this role, though it probably couldn't be used except in large games. But that's fine in its own way I think.

McTsts:
@venomousbirds why did you mark this as a draft?

venomousbirds:
only a few people commented on it and i don't think its ready yet, feel like new roles will need input from more people if they're completely new suggestions.

McTsts:
That's what we're doing right now, isn't it?

venomousbirds:
its been up for more than a month and there's been plenty of people commenting on other PRs at the time

venomousbirds:
hm fair i just thought it wouldnt be worth it to work on, at least for this role book, because i havent seen much activity on it

venomousbirds (approved):

Issue #412, opened 2021-06-25
Lone Wolf => Wolf & Lone Warlock => ??


McTsts:
This was previously discussed but not resolved. 
It was previously suggested that the Lone Wolf turns into a Wolf if they join the pack and that the Lone Warlock turns it a plain non-lycan wolf ("coyote"?)

McTsts:
Everyone has a different opinion on this so we have failed to resolve this issue for v13.

ThomasRubycrust:
I think no option really makes sense for Lone Warlock, because Lone Warlock itself doesn't make any sense. Lone Wolf is a Wolf that's alone and basically becomes a Wolf later on. Lone Warlock is *not* a Warlock that's alone, it's actually exactly the same as Warlock until it joins the pack (so arguably it's *less* lonely than normal Warlock), however the Lone prefix is probably still better to keep than not, because it means LW and LWa have a shared part of their name for their shared ability. I don't think any change of LWa into another role except for Warlock would really seem logical, and changing to Warlock obviously makes no sense. When Lone Wolf changes into Wolf it doesn't really change what the Lone Wolf is (they're still a wolf, just not lonely anymore). Changing the Warlock part of the name doesn't really make much sense either: When the Lone Wolf becomes a Wolf they don't really lose their killing ability - they just share it with the other wolves now, the same can't be said for the Lone Warlock. Even if Lone Warlock was renamed to Lone Something and then turned into a Something when they join the pack, it now wouldn't make any sense for them to lose their ability. Conclusion: none

Shepherdwhen:
[...]

PR #476, opened 2022-05-04
Add Pack Warlock

venomousbirds:
doing it right this time


venomousbirds:
fixes #412

McTsts (approved):

ThomasRubycrust:
Do we want to continue discussing this role or are we dropping it? I'm not sure if the impact on balancing would really be a problem, presumably the PWa would start out being unable to check and then only get a check in the second night of being PW.

venomousbirds:
This was mostly meant as a transformation role like Trickster Wolf, a solution so Lwa doesn't have to turn into a lycan

ThomasRubycrust:
In the issue attached to this Jay brought up that they thought it would be bad because it would remove a key role for smaller games, which is what put this role on hold for now.

venomousbirds:
Little confused by what "it would remove a key role for smaller games" means, where did Jay say this? So I can read it back

ThomasRubycrust:
#412

ThomasRubycrust:
> I understand the whole 'why would a warlock turn into a wolf lol' jig, but I think the role is fine as it is, to be honest. It's a really easy way to give wolves a balanced investigative especially in smaller games where pack might suffer from thinner ranks. The proposed Pack Warlock change would make it much more powerful and would remove what I believe to be a key role for smaller games. There is some current overlap with CF, but CF is imo quite a bit stronger, and so if LW was changed in this, there'd still be that gap. PsyW is meanwhile much weaker.
> 
> Ultimately, names are just names - do they really have to make sense down to the tiniest detail?

venomousbirds:
I still have no clue what "would remove what I believe to be a key role for smaller games" means in this context. Pack warlock is here to give consistency not because of name specifically but to keep Lone Warlock non-lycan. Lone Warlock isn't as strong as normal Warlock, it's only one check every 2 nights

ThomasRubycrust:
I'm also not very convinced that CF is much stronger and PsyW is much weaker. I think LWa as it is can actually be a bit too weak if wolves get unlucky and if they get lucky its more of a warlock. The PWa change would actually make the role a bit more consistent.

ThomasRubycrust:
Rejected by vote

PR #480, opened 2022-05-10
Buff & Add Plague

McTsts:
Make the plague effect immediate. This is much more consistent with other solo effects.


McTsts:
• the rodent not being immediate is unnecessary nerf for the PB that initially only existed to replicate ES behaivor (it was immediate there but the contagious would only start at the start of the next night, the end night contamination achieves roughly the same thing) however changing it to immediate is probably a buff the role deserves
• PD is useless N1 due to the behaivor explained in the previous point

McTsts:
I think plague is ready to be a proper role, but it needs this slight buff / fix. It confused me every time that the effect is end night and not immediate.

Mjnstag:
yep agreed that it should be instant, would make it easier for the plague to spread
👍

venomousbirds (approved):

PR #481, opened 2022-05-11
Fix Hags

### LINKED ISSUE #479: Hags can not be allowed to work as they do right now
They are too easy to control as a voting block. My suggestion would be that hags *must* be wronged once to be able to win.

ThomasRubycrust:
Fixes #479


McTsts (approved):

McTsts:
epic approval from a totally different person

McTsts:
fixed, thanks!

PR #482, opened 2022-05-11
Add Ability Type Explanations to Role Book

ThomasRubycrust:
It just makes sense to have this


ThomasRubycrust:
Is there any type of ability/action I'm missing?

Shepherdwhen:
You added an extra T in targeting and targeted I think, you spelt it targetting and targetted

McTsts:
Turtle's suggestion:

Are you looking for something like this?

> Attack - An attack is an action with the objective to kill a player. Most notably, the wolf pack are given the option to attack one target every night phase. _Killing_ roles possess the ability to attack players.

> Investigation - An investigation is an action with the objective to learn a player's role, alignment or both. _Investigative_ roles possess the ability to investigate players.

> Target - A target is a player that has been chosen by another player with an ability. Targeting happens when a player wants to use their ability, regardless of whether it is triggered once per game, cyclically or upon death, on another player. Targets can generally be changed once per phase with one emergency change in the game, however roles with immediate targeting abilities (e.g. Hooker) may change their target unlimitedly until they lock their target in.

McTsts:
Integrated the naming of the killing/investigative categories like in turtle's suggestion.

cryptidjayden:
communication? like via group roles

McTsts:
oh yeah, maybe "Group Membership" or something would be more accurate than communication (whispers is also communication)

McTsts:
Added a thing for group membership

venomousbirds (approved):

PR #488, opened 2022-05-11
Cupid x LA

McTsts:
If the Look-Alike copies a cupid they should join the cupid


venomousbirds (commented):
Hasn't this happened before? I don't remember what actually happened

venomousbirds (approved):

PR #501, opened 2022-05-12
Improved Mayor Phrasing

ThomasRubycrust:
The mayor role has the information about the double vote twice (and the first time isnt really accurate)


Shepherdwhen:
imo
> the top one or two players being voted

reads slightly better than

> the top one or two players voted

cryptidjayden:
it's "voted on", not just "voted", which i think works much better. i like it

McTsts:
"voted on" is better than "voted"? sounds strange to me

venomousbirds (approved):

PR #502, opened 2022-05-12
More info on how cupid works

ThomasRubycrust:
This is nothing new, just more explanation


cryptidjayden:
i cba to read all of this but i trust u

McTsts:
:thight: I need someone to read it and check if its understandable and stuff

unaligned/align - cupid/cupid alignment
```diff
@@ -0,0 +1,29 @@
+**Cupid Alignment Examples**
```
venomousbirds, line 1: Not all Solos are listed, is this to avoid listing Limited roles?

McTsts: Yeah, I didn't want to list all limited roles, and for a normal game it's probably a bad idea to combine a limited solo with cupid anyway. I only listed Vampire because it has a special edge case, whereas the other limited teams don't.

venomousbirds (approved):

PR #503, opened 2022-05-12
Small Soul Buff

ThomasRubycrust:
Very small buff for soul users, and changing the phrasing a bit so there's not as many repeated "a member of the hell team who can use souls"s


McTsts:
The buff is the last line. People that get attacked several times and hence couldnt be automatically saved, just dont get protected at all.

McTsts:
Do you know if this has ever actually happened? I figured this wouldn't really have much of an impact. It's not supposed to be much of a buff.

cryptidjayden:
i can't remember, it just feels like something we shouldn't add unless it's proven it needs it

venomousbirds (commented):
Interesting, probably won't happen much, probably needs testing because it could decide a game that would otherwise seem over

ThomasRubycrust:
Removed the change from this PR, it's now only an improvement of the phrasing.

PR #506, opened 2022-05-12
Minor RK buff

ThomasRubycrust:


venomousbirds (approved):
Don't think we've ever had this situation but it sounds good

PR #508, opened 2022-05-16
Make Herald ability multi-use

McTsts:


Shepherdwhen:
Idt anything else is needed, just needed clarifying

PR #509, opened 2022-05-16
Clarifiction for how Hooker & Protection works

McTsts:
hooker description specifies that "hooker will not be affected by attacks targetting them", wager protection specifies that "protected from any attacks to target them" => conclusion: if a hooker protects themselves while sleeping with someone they are not protected more than without the protection (since both hooker and protection specifically only protect from attacks targetting them)


cryptidjayden:
yeah

venomousbirds (approved):
Makes sense and clarifies a lot, cuz there were questions and it was generally very unclear

PR #529, opened 2022-05-21
Detective

5uso:
A role whose objective is to find the entire role list.

- During the day can guess like psychic wolf, once per game can instead check like the fortune teller. Affected by strong disguises, not weak.
- During the night they submit to the hosts a role/player list, if it's correct they ascend.

I'm not sure if the role list to be guessed should be the starting one or the current one, that needs to be decidedd and clarified in the description.

The role is kind of "Unaligned Investigative", but that category doesn't exist so Miscellaneous is probably better.

I have no clue of how to balance things so this probably needs quite a few changes


McTsts:
I think I like the idea overall, but it has just occurred to me that this role may potentially really struggle with dead players.

5uso:
Yeah, should probably think of a way to balance that. Maybe instead of the fortune teller once per game they can decide to use a more powerful check on dead people? (they just decide between this and their normal check once per day, both unlimited uses)

McTsts:
So yes/no for the role of an alive player or learn the role of a dead player? Yeah, I think that would probably be better.

5uso:
Sounds about right to balance that, although I'm not sure how it would feel for fun. Maybe having to check dead players could feel like a chore and you don't get that much of a choice. Although many times you could get info by interrogating alive players I guess. Still, I think that having the extra powerful check on alive people is fun but too much if combined with the dead thing if you know what I mean

McTsts:
I think this role could try trading the roles of dead people for doing checks on alive players, maybe.

Mjnstag:
would the detective need to guess the starting role of players or the current role?

McTsts:
> I'm not sure if the role list to be guessed should be the starting one or the current one, that needs to be decidedd and clarified in the description.

McTsts:
We don't need to consider thief.

McTsts:
Updated the checks as discussed before. Also made it so the Detective isn't affected by any effects (disguises, obstructions, redirections), I think them being affected by that would be very counterproductive.

McTsts:
Maybe a slight buff. If their full role list guess is wrong they are informed how many guesses are right/wrong (but not which)

venomousbirds (commented):
Interesting concept
While I do see that it can have effect on the game, I also think it has no reason to be killed and no reason for people to just share some roles with it, there's no drawback. Nothing bad happens when the Detective ascends, it's just there and it's giving me a lot of Survivor vibes. It seems a little too easy

cryptidjayden:
my first thought was "when they correctly guess a player, that player can't use their power the next day/phase/whatever", but that's far too powerful, maybe it could affect votes?

McTsts:
Or we add ice king back instead :)

ThomasRubycrust:
Wolfpack should be able to fairly reasonably claim this role assuming they have a Psychic Wolf (even otherwise its reasonable), though the issue possibly remains that the real Detective might claim to freely and early for this to actually happen

PR #531, opened 2022-05-22
Cupid and hags

McTsts:
Resolve edge cases with cupids and hags


venomousbirds:
assigning myself so i look at this later to read through

venomousbirds (approved):

PR #533, opened 2022-05-22
Bring back Vampire

McTsts:
This PR unlimits the underworld team and contains 3 additional changes related to underworld:

- Exorcist is also unlimited as a counter role. Exorcist has been used in games in the past and seems ready as a full role
- Vampire is re-classified as a "Solo Recruitment" (SR) role, making it so that in partial games it's clear when a role of this type is in the game
- A new limited role: "Devout Villager". It's a Citizen except that it cant lose its role. In games with Underworld it can serve as a replacement for Citizen and is not any more confirmable. However, it's probably a bit nicer to play as DV than Citizen.

Additional notes:
- The werewolf team already has an Underworld counter: the Hellhounds 
- It has occured to me that Underworld team doesn't really have a secondary role like Demon/Plague Carrier/Flute Apprentice, there is technically Undead, but I don't think that should be a starting role. Maybe we can think of a second Underworld main role?


Mjnstag:
> Maybe we can think of a second Underworld main role?

maybe something like a bat that doesnt know the vampire but can spread the demonization if they already visited a demonized person before?

McTsts:
That feels like a worse version of Wisp tbh. I tried looking at the custom role generator to see if that has any interesting abilities, but it doesn't have underworld team 😔

Mjnstag:
maybe a role that knows the vampire and can stay the night at a demonised/undead player and that any visitors would also be demonised?

McTsts:
I like that suggestion more.

Mjnstag:
wait, this would mean that if a ww attacked, then both would die but ww would be demonised and i dont know how to feel about that

McTsts:
Split this into a separate issue because I think that makes more sense.

venomousbirds (approved):

PR #538, opened 2022-05-22
Graveyard Team

McTsts:
This is the old zombie from S1/S2. I dont have the old description but I think it's basically the same.

### DIFF
--- a/limited/zombie
+++ b/limited/zombie
@@ -1,0 +1,7 @@
+**Zombie** | Solo Recruitment - Graveyard Team | Limited
+__Basics__
+Each night, the Zombie may bite another player. If a player is bitten a second time they become a Zombie.
+__Details__
+When a Zombie bites a player a second time, they immediately become a Zombie. The bitten player betrays their team and joins the graveyard channel. They can also immediately bite another player. If they previously were part of another group, they remain in it, but lose their vote. If all members of a certain group have been turned or are dead, their ability will be lost and the group is disbanded.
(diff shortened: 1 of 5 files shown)


venomousbirds (approved):

limited/zombie
```diff
@@ -0,0 +1,13 @@
+**Zombie** | Solo Recruitment - Graveyard Team | Limited
+__Basics__
+Each night, the Zombie may bite another player. If a player is bitten a second time they become a Zombie.
+__Details__
+When a Zombie bites another player they become "bitten". The bitten player does not know about this. When a Zombie bites a player that is already bitten, they immediately become a Zombie. 
+That player betrays their team and joins the graveyard channel. They can also immediately bite another player. If they previously were part of another group, they remain in it, but lose their vote. If all members of a certain group have been turned or are dead, their ability will be lost and the group is disbanded.
```
venomousbirds, line 6: End of this sentence seems a little redundant but idk

McTsts: Not sure what you mean

venomousbirds: > If all members of a certain group have been turned or are dead, their ability will be lost and the group is disbanded.

isnt this just common group knowledge?

McTsts: I think we assume this to be true in general, but I don't think that is defined anywhere. So far this would only happen with Undead and it was part of the Undead role description, so I copied it to here as well.

venomousbirds: in that case we can just leave it here

Issue #536, opened 2022-05-22
Maybe we can think of a second Underworld main role?


McTsts:
> Maybe we can think of a second Underworld main role?

maybe something like a bat that doesnt know the vampire but can spread the demonization if they already visited a demonized person before?

_Originally posted by @Mjnstag in https://github.com/McTsts/Werewolves-Roles/issues/533#issuecomment-1133921070_

McTsts:
That feels like a worse version of Wisp tbh. I tried looking at the custom role generator to see if that has any interesting abilities, but it doesn't have underworld team 😔

_Originally posted by @McTsts in https://github.com/McTsts/Werewolves-Roles/issues/533#issuecomment-1133922668_

McTsts:
maybe a role that knows the vampire and can stay the night at a demonised/undead player and that any visitors would also be demonised?

_Originally posted by @Mjnstag in https://github.com/McTsts/Werewolves-Roles/issues/533#issuecomment-1133923703_

McTsts:
I like that suggestion more.

_Originally posted by @McTsts in https://github.com/McTsts/Werewolves-Roles/issues/533#issuecomment-1133923807_

McTsts:
wait, this would mean that if a ww attacked, then both would die but ww would be demonised and i dont know how to feel about that

_Originally posted by @Mjnstag in https://github.com/McTsts/Werewolves-Roles/issues/533#issuecomment-1133970050_

McTsts:
What do you mean by "both would die"?

Mjnstag:
like when hooker visits someone during the night both die when attacked

McTsts:
Oh I see what you mean

McTsts:
Possibly as a buff that player would just turn into an undead then, maybe? Likey they could be demonized (possibly automatically) as a buff.

Mjnstag:
yea that could work, would also mean that it isnt an ability that they can use without any risks

McTsts:
They'd basically just risk losing the ability, but not losing their life, which I think is probably good.

PR #539, opened 2022-05-22
Add Bat

venomousbirds:
#536


McTsts (approved):

ThomasRubycrust (approved):

ThomasRubycrust:
Suggestion to make haunting end-day to avoid situations where several abilities conflict. You'd always want to use this as early as possible, so I think it's just better as end-day.

PR #543, opened 2022-05-23
Introduce "Start Day" and "Start Night" for Chef and Devil

ThomasRubycrust:
For chef this was relevant last game, for devil this is much less likely to be relevant, but technically the devil can be killed end day (via poisoned food from chef) so it's relevant for them too.


ThomasRubycrust:
Accidentally based this of #542 instead of main, so #542 should be approved first

venomousbirds (approved):

Issue #523, opened 2022-05-21
Remove End Delaying


McTsts:
I say we just remove delaying. It originated from people not pinging for their action and then the hosts asking "are you sure?", the player not responding, making it so that we'd just have to assume it was their final decision at the end of the phase
We can work around that by immediately pinging and asking them if they didn't ping assuming they're still online and if not it just happens

PR #549, opened 2022-05-23
Change End Delaying

ThomasRubycrust:
Implements #523


ThomasRubycrust:
Updated to keep end delaying but make it so you cant abuse it as easily

venomousbirds (approved):

PR #550, opened 2022-05-23
Clarification for Whisper roles

### LINKED ISSUE #510: Devil/Barber x Grandma/Stalker
If the Devil or Barber kill the Grandma or Stalker, do the recipients get kicked out? I'd say they are still only kicked out at the end of the day as usually.

ThomasRubycrust:
Fixes #510 
Am I missing a role that has Whispers?


ThomasRubycrust:
oh I am missing at least one: Chef

ThomasRubycrust:
Stalker & Grandma: No active power, so it continues even after death
Chef & Devil: Both require the role to do something, so they cant happen after death

PR #554, opened 2022-05-23
Change weird Wisp behaivor

### LINKED ISSUE #552: Wisp weird behaivor
If the wisp tries to powder someone its immediate, however if their target is already powdered its an end effect. While this is consistent, it's a bit weird, because technically we cant confirm the wisp's action if it results in igniting and the wisp could use this to realize the person is already powdered and instead powder someone else. I think both abilities should just be end-day.

ThomasRubycrust:
Fixes #552

### DIFF
--- a/other/pyro/wisp
+++ b/other/pyro/wisp
@@ -2,8 +2,8 @@
 __Basics__
 The Wisp visits a person each day, igniting them if they are powdered. 
 __Details__
-The Wisp may visit one person each day, if that person has been powdered, then the Wisp ignites the visited person at the end of the day.
-If that person is not powdered, the Wisp powders that person. This is an immediate action.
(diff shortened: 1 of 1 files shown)


venomousbirds:
imo its more consistent to just leave it the old way and maybe note that the wisp will only be informed if they powdered or ignited at the end of the phase

ThomasRubycrust:
I prefer my solution, but yours works too.

venomousbirds:
Someone else should have a look at it then

McTsts:
> imo its more consistent to just leave it the old way and maybe note that the wisp will only be informed if they powdered or ignited at the end of the phase

Changed to this

venomousbirds (approved):

PR #555, opened 2022-05-23
wisp change

### LINKED ISSUE #552: Wisp weird behaivor

ThomasRubycrust:
Clarify that they do not join pyro channel
Add that they only promote at the start of the next phase - this means that if the last pyro is lynched the wisp action of that day still goes through
Fixes #552


Shepherdwhen:
They probably should join the pyro channel in the case that multiple wisps are left at endgame and they all promote

ThomasRubycrust:
> why wouldnt wisp join the pyro channel?

It wasn't part of the role, and it's what happened in the last game, so I assumed this is the case.

ThomasRubycrust:
We can instead make it so it _does_ join, I only wanted to clarify the current behaivor because I thought it wasnt very clear.

venomousbirds (approved):

PR #564, opened 2022-05-23
Clarify Reaper Redirection

### LINKED ISSUE #558: Reaper & Redirection
When the Reaper picks up the FPA power to redirect, do sleepless players die when targetting the Reaper? Their power will be redirected either way, but I'm unsure if it still counts as using an ability on the reaper

McTsts:
Fixes #558


McTsts:
This is stupid, redirections dont work because the power _fails_ if used on Reaper. so the power just fails and they die... thats it

PR #565, opened 2022-05-23
Make Bat's ability an end-day effect

McTsts:
Make haunting end-day effect to avoid situations where several abilities conflict. You'd always want to use this as early as possible, so I think it's just better as end-day.


venomousbirds:
but its a night effect

venomousbirds (approved):

PR #568, opened 2022-05-23
Wraith Attack consistent with Reaper

McTsts:
should it uh be a SK role now...? 🤔


venomousbirds:
it kinda is? not sure so keep it in SP for now

McTsts:
It's not really a real change to the role so I'm unsure. It's currently SM

venomousbirds (approved):

PR #570, opened 2022-05-23
Update flute player

McTsts:
This is from back when people wouldn't get INDSC, so the FPs would all be asked in the same channel. Now we can instead ask in their own SC and the other FPs wouldnt even know who got asked, so this edge case is irrelevant.


venomousbirds:
do you mean we should only ask in private scs? or ask in group scs and answer in private sc?

McTsts:
I think in games with both role changing and flute players it makes the most sense to ask in private channels.

venomousbirds:
ah

venomousbirds (approved):

PR #571, opened 2022-05-23
Shepherd Reaper Edge Case

McTsts:
I'm crying. Please help.

### DIFF
--- a/limited/reaper abilities
+++ b/limited/reaper abilities
@@ -16,4 +16,4 @@ __Special Cases__
 • Foxes: The wolfpack is not informed of what the Reaper chooses
 • Direwolf: The Reaper may attack one player
 • Exorcist: Makes target sleepless, but still removes other effects
-• Shepherd: Nightmare-tags two players, those players may become sleepless later; Sleepless players work like herded players; Players are informed if they no longer are sleepless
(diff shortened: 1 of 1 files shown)


venomousbirds:
i dont get it

McTsts:
If the Shepherd dies the Reaper gets its power converted to the reapers effect. Usually for roles like FP it is easy and you can replaced "enchanted" with "sleepless". Shepherd has two effects, "tagged" and "herded". This edge cases makes it so by using the Shepherd power the Reaper applies a new effect, "nightmare-tagged" which is the Reaper's equivalent of tagged. If two nighmared-tagged players interact they become nightmared (this is how Shepherd works). The Shepherd has additional poweers such as unherding to steal votes and auto-unherding to escape attacks, the second sentences clarifies that sleepless players can be used in the same way by the Reaper as herded players could be used by the Shepherd. The final sentences adds that nightmared players are informed when they are un-nightmared. This is relevant because unlike herded, nightmared doesn't have a channel, only a message when you get it.

venomousbirds:
maybe only add one shepherd effect, so just tagging to make players sleepless, no need to make it complicated

McTsts:
We can probably do that

McTsts:
PR updated

PR #572, opened 2022-05-23
General Reaper Solo Clarification

McTsts:
help.


venomousbirds:
i dont get this one either

McTsts:
> i dont get this one either

The Reaper gets the powers of the roles that die. FP's power for example is that enchanted players are obstructed. This clarifies that this applies to the Reaper.

limited/reaper abilities
```diff
@@ -7,7 +7,8 @@ __Powers__
 The Reaper is able to use passive powers.
 They are also able to use powers involving a reveal, however nobody will be revealed. 
 Group roles don't give the reaper any powers, nor does the Reaper gain access to any secret channels (e.g. Journal) or info (e.g. Psychopath List, Town Crier) the dead player had.
-Abilities that involve effects (e.g. enchanting, powdering, igniting powdered players) instead work with the sleepless effect.
+Abilities that involve effects (e.g. enchanting, powdering, igniting powdered players) instead work with the sleepless effect. 
+During the two phases after another solo role's death, the powers of that solo's effects are additionally applied to/converted to the sleepless effect/nightmare team.
```
venomousbirds, line 11: i just dont get this sentence, i dont understand what it means so its probably worded weirdly

McTsts: Examples:
- If a FP dies the Reaper gains the FP powers for two phases.
- The FP powers are:

1. Enchanting people immediately

2.  Enchanted people are obstructed

> Abilities that involve effects (e.g. enchanting, powdering, igniting powdered players) instead work with the sleepless effect.

Solves 1) - Instead of enchanting the Reaper gives people nightmares

Phrasing a general solution sentence for 2) is very difficult, so this was my best attempt and we can probably improve it.

> During the two phases after another solo role's death, the powers of that solo's effects are additionally applied to/converted to the sleepless effect/nightmare team. 

> During the two phases after another solo role's death

When a role like FP dies their powers are used by the Reaper for the next two phases

> the powers of that solo's effects

Powers that that solo's effect has. For FP that would be the powers of the "enchanted" effect, i.e. obstruction

> are additionally applied to/converted to the sleepless effect/nightmare team. 

Those powers are converted to fit the nightmare team. So "enchanted" turns into "nightmare" and the obstruction showing "Flute Player" turns into "Reaper"

venomousbirds (changes requested):
The sentence needs rephrasing, I only understand it because of the explanation, but I don't think anyone else will without it

McTsts:
I replaced the sentence entirely and provided an example. Is this better?

venomousbirds (approved):

PR #574, opened 2022-05-23
Herding Dog Buff

venomousbirds:
Maybe need to add more but this is proposed

### DIFF
--- a/limited/herding dog
+++ b/limited/herding dog
@@ -2,5 +2,6 @@
 __Basics__
 The Herding Dog wins with the shepherd and joins a channel with the shepherd to communicate securely.
 __Details__
-During the day, the Herding Dog can un-herd a herded player to steal that player's vote. During the night, if the Herding Dog is attacked, a herded player will be automatically unherded to protect the Herding Dog.
+During the day, the Herding Dog can un-herd a herded player to steal that player's vote. Players with negative votes can't have their negative votes stolen (but will still be unherded).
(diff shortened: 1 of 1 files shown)


Shepherdwhen:
The thing is Hearding Dog was never intended for a list, it was just in case of a LA on Shepherd, I had thought of buffs however, one being that each night the Hearding Dog visits a player and acts as a faux tag, if a tagged player visits the HD's target, the HD's target is Hearded

McTsts:
> The thing is Hearding Dog was never intended for a list, it was just in case of a LA on Shepherd, I had thought of buffs however, one being that each night the Hearding Dog visits a player and acts as a faux tag, if a tagged player visits the HD's target, the HD's target is Hearded

That's basically #539

McTsts (approved):

Shepherdwhen:
> > The thing is Hearding Dog was never intended for a list, it was just in case of a LA on Shepherd, I had thought of buffs however, one being that each night the Hearding Dog visits a player and acts as a faux tag, if a tagged player visits the HD's target, the HD's target is Hearded
> 
> That's basically #539

Kinda but instead the visitor is only herded if they're already tagged, and it acts as a temporary tagging to the HD's Target instead

Shepherdwhen:
The other Idea I had is they passively learn the role and name of the visitor who caused the herding, this would also make the HD a Solo Investigative

McTsts:
Investigative is a sub-category of Power and does not necessarily need to exist

PR #584, opened 2022-05-24
Change Herding Dog

McTsts:
- 1 unherding instead of 2 makes more sense IMO. HD alone is pretty weak.
- Protecting the Shepherd should be the higher priority


Shepherdwhen:
Just made a change @McTsts

McTsts:
Oh yeah, you're right. Updated the category files as well.

venomousbirds (approved):

PR #585, opened 2022-05-24
Create Tracking Wolf

Shepherdwhen:
A new pack WI idea, can't think of a name rn which is why it's called "? Wolf"


Shepherdwhen:
It was just a basic draft but yeah I agree with these changes

McTsts:
Overall I think the idea could work. It's similar to AT, but also unique enough. I don't think it would be overpowered because categories aren't always enough info (a Huntress and Assassin would be treated very differently. Bard/Bartender and Citizen are completely different levels of strength), but still, it's enough info to be very useful.

McTsts:
Rewrote the role according to my feedback.

ThomasRubycrust:
@Shepherdwhen check the new description at some point when you have time, and let me know what you think.

Shepherdwhen:
The new changes seem fine, for the name Stalking/Stalker Wolf would work if we didn't already have a town role called stalker

ThomasRubycrust:
We'll never add Stalker Wolf, but I still don't think it's a role name we can use.

cryptidjayden:
i like this! can't think of a name, though, which is a shame bc stalker would fit so well.

McTsts:
I'm kinda opposed to just removing Stalker, but I think renaming this one to Stalker Wolf and revamping Stalker under a new name is fine, and Stalker is very rarely used (last used 22 games ago, only a total of 3 uses on WWR) so it probably wouldn't be too confusing.

cryptidjayden:
i'd be fine with that tbh

cryptidjayden:
gonna link back to #599 to discuss the stalker thing more bc i had another thought

venomousbirds:
i dislike this, i think its a really weird an unneccesary check

venomousbirds:
clarification on my comment earlier

the check is unusual and unneccesary, it just adds another weird check mechanic to the game and i dont think we need that. 
all the kinds of threats are just a little vague and i think it will add more confusion than clarification. overall i think this is a pretty bad WI and it won't help much
also i dont think stalker wolf makes sense as a name. no need to delete another role for a role draft.

ThomasRubycrust:
The check works the same way as the AT check

Shepherdwhen:
Also the Check isn't confusing, the higher the threat the more rewarding it is to have that player dead most of the time

Shepherdwhen:
The only time it isn't is weird edgecases

venomousbirds:
no this check has more than the AT check and i dislike that, it only complicates the game

>Check isn't confusing

i find it to be confusing, its put weirdly imo

venomousbirds:
its a big bundle of words, i think putting it like the first commit (new line for every sort of check) would make it more clear

ThomasRubycrust:
Honestly, I find this much much more readable than the initial more table like list.

Shepherdwhen:
> Honestly, I find this much much more readable than the initial more table like list.

I personally find both readable but the current one fits the consistency of other roles

cryptidjayden:
i like the role! i like the more dynamic threat interactions and i feel pack could really benefit from a WI that is (imo) stronger than PsyW. i also saw vera's point about the mildly confusing bundle of words; to fix this, i made sure each scenario was phrased in the same way, which i hope helps. (i also changed the last line to be consistent with aura teller.)

a thought: should we implement the different threat levels with aura teller? it runs the risk of making AT more/too powerful, but could add more interest to what is (imo) a more "boring" role than many of the other TIs. it'd also help differentiate from huntsman somewhat. if anyone wants to expand on this further, a new issue or PR could be created.

McTsts:
I think the phrasing is now better. I don't think AT should be changed.

Shepherdwhen:
we could also call this role prowler wolf/hunting wolf/tracker wolf if we're not removing stalker

McTsts:
I think those honestly make more sense than Stalker Wolf? Those seem more like things a wolf would actually do.

venomousbirds:
I think Tracker Wolf fits best out of those

McTsts:
Updated role to `Tracker Wolf`

venomousbirds:
still dont like it tho, but that name does fits best

cryptidjayden:
i think i prefer prowling to tracker, but either work!

venomousbirds:
Tracking wolf? cuz its also infecting wolf

McTsts:
Tracking Wolf also works, but for Infecting the other option is Infected which doesnt make any sense.

venomousbirds:
true
i do think tracking sounds a little better than tracker

ThomasRubycrust (approved):

venomousbirds (approved):

PR #589, opened 2022-05-25
Improve Undead Phrasing

ThomasRubycrust:
I think the role description for Undead is really bad right now. It has redundancy, a bit too much flavor and some things that aren't even really true. There's lots of leftovers from the style role descriptions had in the pre-WWR role book versions. It seems like more things just got added to the role description without really trying to keep the role description simple and readable
I've moved the flavor to the very start, where I think it makes the most sense. Additionally, I've gotten rid of some sections, combined some others and overall tried to make the role description better. There are no changes to how Undead works.

### DIFF
--- a/other/underworld/undead
+++ b/other/underworld/undead
@@ -1,8 +1,7 @@
 **Undead** | Solo Miscellaneous - Underworld Team
 __Basics__
-The Undead are the vampire’s secret army. If a demonized player were to die during the night, they will instead lose their original role and become an Undead.
+The Undead is a dead player who pretends to be alive, and they are part of the vampire’s secret army. If a demonized player dies during the night, they will not die, and will instead lose their original role and become an Undead.
 __Details__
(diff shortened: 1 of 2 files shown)


other/underworld/undead
```diff
@@ -1,8 +1,7 @@
 **Undead** | Solo Miscellaneous - Underworld Team
 __Basics__
-The Undead are the vampire’s secret army. If a demonized player were to die during the night, they will instead lose their original role and become an Undead.
+The Undead is a dead player who pretends to be alive, and they are part of the vampire’s secret army. If a demonized player dies during the night, they will not die, and will instead lose their original role and become an Undead.
```
venomousbirds, line 3: phrasing here makes it seem like the player is actually a dead participant, and not just a participant who had their role changed to Undead by getting attacked while demonized

McTsts: this is actually just from the original description (just moved to basics instead of details), however we can change it.

venomousbirds (changes requested):
Everything except the Undead Basics is fine, might do better without it or with the old one

McTsts:
What about this? Keeps the pretending part in a way that I think makes more sense

venomousbirds (approved):

PR #591, opened 2022-05-25
Some changes to cat

ThomasRubycrust:
So I was reading the Reporter info, and I saw it mention Dog and then it occured to me that Dog has far more potential than I would've thought. Basically with a Dog you would be unable to know whether lynching a dog was a mislynch or not. However, of course with Dog the issue is balance. Cat's entire purpose is to address this issue, so I think games with both Cat and Dog would probably work out, since balance is still ensured.

Dog has four options:
Town Crier
Cursed Civilian
Wolf
Citizen

Town Crier and Wolf are considered to be opposite, so Cat just switches them. CC is considered to be inbetween those two (can benefit both alignments), so it makes sense that cat would become a Psycho, a role that in some way wants people from both alignments dead. That just leaves Citizen. The opposite of Citizen can't be Wolf because that's the opposite of TC, a role clearly stronger than Citizen, so I assume that's why the original Cat has Survivor. But Citizen is still definitely town-sided, so Survivor isn't a particularly good opposite _and_ more importantly Survivor doesn't exist. Packless Wolf would probably be the best equivalent, but it sucks and doesn't exist either. I think Angel would make for a better opposite - it exists and is generally a more wolf sided role, but not nearly as much as a wolf.

Also improved phrasing.

### DIFF
--- a/categories/Unaligned Align
+++ b/categories/Unaligned Align
@@ -1,11 +1,11 @@
 **Unaligned Align**
 Dog
+Cat
 Look-Alike
 Cupid
 Several Cupid Variants (`$i uac`)
 
 *Apprentice*
-*Cat*
 *Thief*
 *Hag*
 *Several Hag Variants (`$i uah`)*
--- a/limited/cat
+++ b/limited/cat
@@ -3,5 +3,5 @@ __Basics__
 During Night 1, the Cat will immediately become the opposite role dog choses for the rest of the game. 
 __Details__
 If the dog chooses town crier, the Cat will turn into wolf; cursed civilian will turn the Cat into a psychopath with the dog in the list; and wolf will turn the Cat into a town crier.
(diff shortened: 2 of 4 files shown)


Shepherdwhen:
Wouldn't the opposite of TC be KW as they both start with a little bit of info

cryptidjayden:
what if dog just couldn't turn into cit? there's no reason a dog would ever willingly let itself turn into cit, and it only happens when somebody genuinely forgets. instead, why can't they just turn into cursed civ if they don't submit a role?

ThomasRubycrust:
> what if dog just couldn't turn into cit? there's no reason a dog would ever willingly let itself turn into cit, and it only happens when somebody genuinely forgets. instead, why can't they just turn into cursed civ if they don't submit a role?

sure, lets just do that

cryptidjayden (approved):

venomousbirds (approved):

PR #592, opened 2022-05-25
Delete corrupt wolf

ThomasRubycrust:
I just don't think this is a very interesting role, there's really not much you can do with it and in general it's just annoying for town. I'd rather buff Saboteur Wolf or something like that, Saboteur is a far more interesting power imo.

### DIFF
--- a/categories/Custom Lycans
+++ b/categories/Custom Lycans
@@ -15,7 +15,6 @@ Cupid Wolf
 
 White Werewolf
 
-*Corrupt Wolf*
 *Knowledge Wolf*
 *Direwolf*
 
--- a/categories/Custom Wolfpack
+++ b/categories/Custom Wolfpack
@@ -13,7 +13,6 @@ Sneaking Wolf
 White Werewolf
 
 *Pack Tanner*
-*Corrupt Wolf*
 *Knowledge Wolf*
 *Direwolf*
 
--- a/categories/Werewolf Power
+++ b/categories/Werewolf Power
@@ -4,5 +4,4 @@ Infecting Wolf
 Saboteur Wolf
 Tanner
 
-*Corrupt Wolf*
 *Pack Tanner*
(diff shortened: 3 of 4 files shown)


cryptidjayden:
we hate corruption here at werewolves revamped.

venomousbirds (changes requested):
okay so i missed it
why do we wanna delete this? i dont seen anything wrong with it
we could make it a pack WI

venomousbirds:
if we let it cancel the election, it can then get the role of the top-voted player

ThomasRubycrust:
> okay so i missed it why do we wanna delete this? i dont seen anything wrong with it we could make it a pack WI

Election cancelling is just overpowered with how short our games usually are. If we had 20 day long games like in the past this would be completely fine. 

> if we let it cancel the election, it can then get the role of the top-voted player

The role already is overpowered.

> we could make it a multi-use ability in the same way the www kill is multi-use, or maybe even less often

that just makes it worse

Shepherdwhen:
If you don't want to delete it we can always make it stop the top voted player getting the role and instead the second most voted player gets it

venomousbirds:
i honestly cant remember corrupt wolf cancelling a good election except for maybe 1 time, so i think getting it to learn the top players role (keep it a one time use) isnt all too bad

ThomasRubycrust:
> i honestly cant remember corrupt wolf cancelling a good election except for maybe 1 time, so i think getting it to learn the top players role (keep it a one time use) isnt all too bad

I mean CW should always just cancel the D1 election to offset Mayor/Reporter by at least a day.

cryptidjayden:
honestly i can only remember CW being used in that one VIP game. ts is right, it's already very powerful, and honestly pretty boring: just delays town getting reporter by one day, essentially. ts is right, if our average games ran longer than 4 or 5 days (or if we had more elections), maybe it'd be more interesting. i'm not necessarily for or against deleting it, but i wouldn't use it personally, and i wouldn't mind if it were deleted.

McTsts:
Honestly same, if you really want to keep it, then I guess we can keep it? But it just seems like a boring role that's always guaranteed to do the same and I'd personally never use it.

ThomasRubycrust:
@venomousbirds please merge or close/delete this PR

cryptidjayden:
@venomousbirds

PR #594, opened 2022-05-25
Bring back ice royals

ThomasRubycrust:
#529 made me want to give Ice Royals another try.

I've always wanted the Ice Royals to work, but I gave up on them previously. The Detective PR (and the issue that nothing would really stop it) brought this team back into my mind. Here's some changes that I think would make this a role/team that can work again:

- The Ice King doesn't gain votes. I think this would rarely be a huge advantage (it does hide how many people are frozen if everyone votes the same way), but it's just too easy to use the voting information to figure out the IK
- Allow the IK to guess the same player several times (e.g. they might deduce someone is TI, but unsure which of the two - now they could just do both guesses)
- Give the IK feedback for their guesses - I believe this was already done, but it's now clarified
- Add an obstruction - In issue with IK was that it had no disguising or protective abilities. I thought about it a bit and I think here an obstruction would make quite a bit of sense. If the IK does more risky plays (to figure out more roles), they're more likely to be checked, but in especially those situations they're also more likely to be able to make use of their obstruction. 
- Dont bring back Ice Queen - I don't think IQ was a good addition to this team. The any role imitation feature works better as a separate team (nightmare), so I'm not planning to bring this power back. I think an investigative based secondary role would make more sense here.

- The new IQ: I'm unsure if this is good, but since the Detective PR made me think of IRs, I want to give the secondary role the power the Detective would have. The Detective can check people (same here) and talk to people to figure out info. The IQ obviously cant reveal to figure out roles, so I thought that maybe a whispers ability could work. Some people might be convinced to share info, in exchange for info from the IQ (idk if this would work, I dont think we really have something intended to work like this at all yet)
Promotion is lame imo, so instead the IQ just gets a freezing power if IK dies. It's much weaker than IK's freezing, but it's something.  
And IK IQ team can (almost) guarantee one freeze per night, which I think is decent.

EDIT:

- Ice Prince - Basically an imp like role that is created when an IR is copied, it's only power is having the obstruction


Shepherdwhen:
> The Frozen thing on IQ should that be applied to some other solo secondary roles like Wraith and Herding Dog

That's confusing I mean as in secondary solos have some way to win if their main solo is killed off

ThomasRubycrust:
**Promotes**
Flute Apprentice
Wisp
Plague Carrier
Corpse
**Has Powers Itself**
Bat
Demon
Herding Dog
**Gains Additional Powers**
Ice Queen

**kinda nothing**
Wraith

Shepherdwhen:
I forgot We were buffing HD

Shepherdwhen:
Also speaking of nightmare team would reaper get the ability of IK but just have players be nightmares instead? Same as IQ would the person targeted and put in the channel be nightmared

ThomasRubycrust:
> Also speaking of nightmare team would reaper get the ability of IK but just have players be nightmares instead? Same as IQ would the person targeted and put in the channel be nightmared

I suppose for IK, it could do X guesses and right guesses would become sleepless. IQ would depend on whether IK is alive. If they are Reaper only gets the check + channel, but not sleepless player. Both IK and IQ would enable an obstruction for 2 phases, which would hide the Reaper.

ThomasRubycrust:
Note: If this idea isn't shut down by another Admin, an Ice Prince or similiar sort of role would also be needed. As a whispers role IQ cant be copied. If IQ cant be copied I feel like IK should also be uncopyable.

ThomasRubycrust:
> > Note: If this idea isn't shut down by another Admin, an Ice Prince or similiar sort of role would also be needed. As a whispers role IQ cant be copied. If IQ cant be copied I feel like IK should also be uncopyable.
> 
> Is it whisper roles or whisper solos only?

All roles with personal whispers (Devil, Stalker, Grandma, Chef) can't be fully copied. This is resolved in one of two ways: "Special Copy", the LA becomes a different role instead, or "Join", the LA joins that role and they share the same power (technically the role is copied, however the power isn't).
Flute Player whispers differs from IQ and Devil whispers - it isn't personal. Since all FPs share the same whispers anyway copying a FP causes no issues.

ThomasRubycrust:
The rationale behind this is that for LA we simply rename their channel and pin a new role description, they then join any additional channels the copied role might have (e.g. wolfpack, hell, cult, etc). For something like Journalist this would mean they join the journal channel / for Chef they would join chef_talk (and nothing else). This is fine and we could let the LA and the copied role make a cc on their own so (since talking in journal or chef_talk isnt very convenient), but simply having them join together into a main SC just makes more sense.

ThomasRubycrust:
@venomousbirds @cryptidjayden is either of you particularly opposed to giving Ice King another try?

McTsts:
Is that a no...?

venomousbirds:
i'm gonna have to read through all the files because i honestly skipped them because i didn't think it was worth a read

venomousbirds:
i dont get why theres an ice queen (old) but not an ice king (old) or why ice queen (old) is archived but other files are deleted

McTsts:
> i dont get why theres an ice queen (old) but not an ice king (old) or why ice queen (old) is archived but other files are deleted

The new ice queen is a completely new role, whereas ice king isn't. When a file is edited it too much github doesn't detect it as moving and instead deletes the old file and creates a new one.

McTsts:
> i'm gonna have to read through all the files because i honestly skipped them because i didn't think it was worth a read

You can read the initial PR message.

McTsts:
> so which of these files are worth checking, if ice queen (old) will be archived for example, i dont think its worth reading/checking

You should read the initial PR message and the "ice king" and "ice queen" files. Ice queen old is simply a rename of the old ice queen so that the new one could take its place.

ThomasRubycrust:
Check the message at the very top for a summary, or just these files:

https://github.com/McTsts/Werewolves-Roles/blob/fc660f595f2bb32d9ab12dd0dee1f031f3d240ab/limited/ice%20king#L1-L14
https://github.com/McTsts/Werewolves-Roles/blob/fc660f595f2bb32d9ab12dd0dee1f031f3d240ab/limited/ice%20queen#L1-L15

cryptidjayden:
i like both roles in theory; in practice, IK seems very hard without IQ, even in a public list (in a partial it would most likely be completely unplayable), which restricts its use to larger rolelists, whilst IQ is clearly designed to only be played with IK ingame.

cryptidjayden:
(which means its use is also restricted to larger games. being designed around another role is not an inherently bad thing, obvs, otherwise we wouldnt have solo secondaries)

McTsts:
I think recently we've been doing two player solo teams for a huge percentage of games, and I think in general that's better, ofc it doesn't work for all games, but I don't think it would be a big issue for ice royals to always be 2 players

Mjnstag:
In that case we could maybe add a lesser Ice Royals role like ice servant or something who can guess at peoples role category to narrow it down for the ice king in a more closed game, wouldnt freeze them obviously

cryptidjayden:
ts has a point, but i like marten's fix idea

McTsts:
I don't get it? Doesn't that make it harder?

cryptidjayden:
how so? call it ice duchess or ice butler or something, give it an invest which returns role category rather than role itself. perfect tertiary role for if we have ice royals in a game with more than 20 or so players, i think

McTsts:
Oh you want to have that for a 3p team, I see.

cryptidjayden:
oui oui

cryptidjayden:
haha ouizer

McTsts:
I'm not sure if 3p solo teams is something we really do currently? currently our games tend to be in the <20p range and I think a 3p solo team is too much there

cryptidjayden:
the theoretical fourth role could also work in <16p lists, where ik+iq might be a bit too much

McTsts:
That's what I thought the suggestion initially was for

PR #597, opened 2022-05-25
Thief switch disguises

McTsts:
Thief should also disguises that belong to a role (Devil) but not ones like Tanner


venomousbirds:
wdym exactly?

McTsts:
If a Thief switches with a Devil they should also get the disguise as it belongs to devil and not the player. If they switch with a player that was tanned they shouldnt get the disguise, as it belongs to the player and not the role.

venomousbirds (approved):

PR #617, opened 2022-06-01
Change Reaper to SK

### LINKED ISSUE #603: Nightmare Team inconsistency
> wraith is SK but reaper isnt

@McTsts

ThomasRubycrust:
Fixes #603

### DIFF
--- a/categories/Solo Killing
+++ b/categories/Solo Killing
@@ -6,6 +6,7 @@ Wisp (Pyro Team)
 White Werewolf
 Plague Bearer (Plague Team)
 
+*Reaper (Nightmare Team)*
 *Wraith (Nightmare Team)*
 *Bitter Hag (Hag Team)*
 *Scared White Werewolf*
--- a/categories/Solo Power
+++ b/categories/Solo Power
@@ -2,7 +2,6 @@
 Flute Apprentice (Flute Team)
 Flute Player (Flute Team)
 
-*Reaper (Nightmare Team)*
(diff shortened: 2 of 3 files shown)


cryptidjayden (approved):

venomousbirds:
not sure if this is neccesary, its more power than killing

McTsts:
> not sure if this is neccesary, its more power than killing

I agree, however with Wraith being SK I'd find it weird if this wasn't...

venomousbirds:
we have teams in different categories with other alignments, so it's not inconsistent (example im taking is cm-cl so take that for what you will)
i think its better to focus on the correct categorization

McTsts:
> we have teams in different categories with other alignments, so it's not inconsistent (example im taking is cm-cl so take that for what you will) i think its better to focus on the correct categorization

That misses the point. Wraith and Reaper share their power so they should also share their categorization.

McTsts:
I'm unsure what to do about this. Reaper does feel more like a Solo Power, but its strange with it having the exact power of a Solo Killing role.

venomousbirds:
i do think SP fits a lot better
wraith is no doubt a SK, it's only ability is killing sleepless players when they attack the Wraith
reaper has many more abilities that don't include killing, killing here isn't the main effect at all, just a passive 'protection'
the focus is way more on using the powers of the dead players, making it a SP
SK doesn't really fit

McTsts:
I don't like it, but I think keeping it as SP is probably better.

PR #621, opened 2022-06-01
Remove Stalker :(

### LINKED ISSUE #599: Buff Stalker
stalker lowkey fucking sucks balls. it's easy to confirm and communication is good, but incredibly weak outside of that, and as such never gets used - it's kind of like a weaker grandma. either scrap the role or buff it. a name change would also free up stalker to be used for the proposed ? WI role

ThomasRubycrust:
Fixes #599 

:(

### DIFF
--- a/archive/stalker
+++ b/archive/stalker
@@ -1,4 +1,4 @@
-**Stalker** | Townsfolk Power
+**Stalker** | Townsfolk Power | Archived
 __Basics__
 Each night, the Stalker may assign two extra votes to a player in the next day’s polls.
 __Details__


ThomasRubycrust:
If we have an idea to make it more unique and useful, which right now I don't :(

Shepherdwhen:
We need the Name rn for another role

venomousbirds:
theres a whole lot of other roles that are hardly used and dont get to be thrown out
grandma and stalker are different voting powers, both unique and i think it's not neccesary to remove one
saying its gotta be removed because of name is the worst argument, especially because the role it needs the name for isnt approved

Shepherdwhen:
No one can think of a buff

venomousbirds:
and that means we gotta delete the role?

PR #622, opened 2022-06-01
Wager: Allow changing attack target

### LINKED ISSUE #611: Wager change
> devils wager: we should do is add a thing that wager recipient can change, but cant have their soul back

@McTsts

ThomasRubycrust:
Fixes #611


ThomasRubycrust:
Added that, I left it out initially, because I want to avoid adding even more text, but turns out it doesnt require many words.

cryptidjayden (approved):
change the however in line 18 to a but, good change

cryptidjayden:
fuck yeah my github skills are fucking elite

venomousbirds:
pro gamer move right there

PR #623, opened 2022-06-01
Tell PC how many people have the plague

### LINKED ISSUE #615: Plague should know number of contaminated people
> It's basically just a counter in the role channel that is updated (editing the message) to reflect the current number of infected alive people. I think it could be interesting if either the plague doctor or the plague carrier had something like that (not the bearer tho)

-suso

_Originally posted by @venomousbirds in https://github.com/McTsts/Werewolves-Roles/issues/606#issuecomment-1142567437_

ThomasRubycrust:
Fixes #615

I put it as start-phase because I think a counter is strange and end-phase would conflict with kills.


cryptidjayden (approved):
needs a comma after phase, good change

venomousbirds (approved):

PR #632, opened 2022-06-02
Remove pyronner

ThomasRubycrust:


venomousbirds:
wait why would this be removed but scwww stays?

McTsts:
scwww seems slightly more serious, but we can remove both

venomousbirds:
i like both tbh, pyronner could be used in normal games too because a lone pyro can be incredibly difficult

McTsts:
I think it needs a different name then

12Metatron:
I agree, this could be a useful role to keep, but it should get a better name for sure. Maybe scared pyromancer, idk just a thought

McTsts:
I think a better buffed pyromancer would be something that involves powdering in some way, e.g. immune against attacks of powdered people, so I think we should still remove this and add another buffed pyro instead.

venomousbirds:
that would work

venomousbirds:
didnt mean to not accept it

McTsts:
@venomousbirds why did you close this then

venomousbirds:
i meant to accept it but fucked up lmao

venomousbirds (approved):

Issue #633, opened 2022-06-02
WWW Team


ThomasRubycrust:
Have we ever considered just having WWW be part of the white wolf team or something? RN if a WWW is copied there's two WWWs, which just sucks for everyone, the wolves, the original WWW and the new WWW. We could instead just let them join and have them keep the same amount of kills as a single WWW. Seems way more balanced.

ThomasRubycrust:
The only downside I see is, that then we can't have two WWWs at the same time, which idk if we do that, like ever. And rn its just inconsistent that WWW is the only solo without a team.

Shepherdwhen:
I had an idea for a non pack who can win solo but also wins if www wins

The silver fox

Each day they target a player, if the player is pack they're attacked, if not they're not attacked

The WWW(s) learn who the silver fox is attacking

The silver fox mimics a fox in the game

ThomasRubycrust:
This just seems like a pretty bad suggestion overall:
- If it was a WWW Copy Case that teams with the WWW they'd just inform the WWW (learning all pack members) meaning its just a packless/buffed WWW
- Even without that it still increases the amount of WWW kills (which isnt particularly good), with the limitation being that you hope they don't actually figure out who's in the pack (a pretty bad limitation)
- A packless wolf doesnt really have any hope of distinguishing between a WWW or normal Wolf meaning this just encourages team killing

venomousbirds:
I'm not sure if www needs a team
but the issue that a www can be copied (meaning a new www is created who knows the other www and can easily target that one with its kill)
maybe just straight up let them join the channel but still only give them 1 kill? would just be a lookalike case?

ThomasRubycrust:
> I'm not sure if www needs a team but the issue that a www can be copied (meaning a new www is created who knows the other www and can easily target that one with its kill) maybe just straight up let them join the channel but still only give them 1 kill? would just be a lookalike case?

This is exactly the suggestion. If they join together however they are a team, so we might as well define WWW as having a team (instead of it being the only teamless Solo) to begin with.

venomousbirds:
well its only an edge case, its not supposed to have a teammate at the start of the game (for example, putting 2 wwws would still make it 2 solo teams) 
i do think www being completely solo adds more to the game than _yet another solo team_

ThomasRubycrust:
[...]

PR #634, opened 2022-06-02
White wolves

ThomasRubycrust:
Implements #633


venomousbirds (approved):
i dont think this is neccesary at all but nobody else seems to wanna say anything about it so i guess we can merge it

PR #635, opened 2022-06-02
Delete the two s0g0 alt Vampire/Devil teams

ThomasRubycrust:


cryptidjayden:
can't we just delete them? we're never going to use them again

ThomasRubycrust:
actually, we probably can. Some roles are worth keeping in the archive, but these dont even really have a description, so might as well delete them

ThomasRubycrust:
Updated PR to delete both that entire team as well as the entire Heaven team

cryptidjayden (approved):

PR #672, opened 2022-06-11
Create Amulet Holder

McTsts:
A rewrite/rework of amulet holder:
image

I think this is actually a fairly unique role and an interesting way to have a protective role that isn't too strong. I think there are a lot of ways the amulet could be relevant to the game. For example, a wolf may get it, but must now decide whether to pass it back to another townie or to a wolf.


ThomasRubycrust:
Should the amulet persist if the Amulet Holder dies or should it disappear/shatter

venomousbirds:
shatter would be interesting, as in attacking or killing the current holder

ThomasRubycrust:
Yeah, I think that would be interesting, though then Amulet Holder is a kinda bad role to play as

venomousbirds:
it doesn't sound like a very towny mechanic to random kill whenever the amulet holder is killed
but it also kind of does?

ThomasRubycrust:
I think it would make sense with the way the role works so far, however I just think it would make it so AH is very bad too play. You basically get no powers, but if you die another person may also die.

cryptidjayden:
there may be a couple of issues in terms of how similar "amulet holder" and "holder of the amulet" are. this could potentially be solved by not using the word holder in the role name (e.g "amulet bearer") or giving it an entirely different name (e.g "cleric")

ThomasRubycrust:
Good point. Amulet Bearer is a bit too make like Plague Bearer in my opinion and I'm just not a fan of cleric. What about something along the lines of "Jeweler", since the most defining part of the role is that it adds/creates the Amulet, not that it 'holds' it.

cryptidjayden:
jeweler doesn't seem quite right to me personally, not a massive fan - an amulet is (at least to me) far more of a religious symbol than it is a piece of jewellery. are you not a fan of cleric as an individual word or because you don't think a religious word is right? @ThomasRubycrust

McTsts:
Both tbh

McTsts:
I'd see the amulet more as a piece of magic than anything religious in nature. And I just dont like Cleric.

cryptidjayden:
disliking cleric is perfectly fine by me. i was gonna suggest acolyte but i'll shelve that for an upcoming solo suggestion i'm working on. maybe wizard, to tie in with witch?

McTsts:
Doesn't fit at all imo. This role barely does anything actively, and if I hear Wizard I definitely would assume they did do something.

McTsts:
A strange alternative would be for Amulet Holder to not even be a role and it being more of a !_Target type of thing, that just happens to someone at the start - would solve the issue that AH is a pretty lame role by itself.

cryptidjayden:
i really like that option.

cryptidjayden:
@McTsts yeah i think that's the neatest solution, do you wanna go ahead and change the files?

McTsts:
I think I'd like to wait for @venomousbirds's opinion before we introduce a basically new type of mechanic + we need a name for that and a place to put it

cryptidjayden:
fair nough

venomousbirds:
So from what I currently read, it's a role that you start out with but it's powerless once the amulet is given to another person. That in itself is interesting, kind of how Fletcher works (except Fletcher can still do things once they've selected a customer), so for this role itself I'd suggest a power like Fletcher has (instead of taking away the huntress ability, maybe shatter the amulet to kill whoever holds it at that moment)

For making it into what's basically a modifier, I'm not too sure how that would work out. It seems a little random to me. What decides who will be the first to hold the amulet? I think it's a little too uncertain and it would work better as a role.

ThomasRubycrust:
That's a good alternative suggestion I think. Solves the problem of the role being too boring, without needing to do anything new. Unsure how well any of this would work in actual games, but it sounds good at least.

venomousbirds (approved):
More like fletcher, would be an interesting mechanic ingame

ThomasRubycrust:
This would actually turn the role into a TK role.

cryptidjayden:
brings back the problem of the name though. now it's more active, we could reconsider wizard?

McTsts:
Still think thats bad for the same reason. This role _really_ just doesnt feel like one where the name Wizard fits

McTsts:
I mean to be honest if you choose one hour before phase change, and then go offline that's kind of on you

McTsts:
We could allow the players to provide a list of players to which they want to pass the amulet

McTsts:
And it picks the first available one

McTsts:
>  **Mr. Turtle**: Or you could just do a hard limit of 3 players in order and you can't resubmit or something.

I like that idea. We should do that.

McTsts:
>  **Mr. Turtle**: Or you could just do a hard limit of 3 players in order and you can't resubmit or something.

Did this.

McTsts:
Added that and rewrote the description a bit to make it shorter:

**Amulet Holder** | Townsfolk Killing | Limited
__Actions__
At the start of the game, the Amulet Holder is the first holder of the *Amulet of Protection*. Whoever holds the amulet is protected from attacks during the night, but must pass it to another player during the day.
__Basics__
When passing the amulet, up to three players may be specified (either at once or separately) and the amulet will be passed to the first player who has not previously had the amulet. Selecting a target is an immediate effect and should the selection fail the holder is immediately informed, however passing the amulet is an end-day ability. If the holder of the amulet does not or cannot choose a player to pass the amulet to, it shatters at the end of the day, killing the holder.
Holders of the amulet do not know who gave them the amulet, nor do they know who has previously held it.
The Amulet Holder must pass the amulet during Day 1.
Once per game, the Amulet Holder may choose to shatter the amulet during the day, destroying it and immediately attacking the holder. 

__Simplified__
At the start of the game, the Amulet Holder is the first holder of the *Amulet of Protection*. Whoever holds the amulet is protected from attacks during the night, but must pass it to another player on the following day. Each player can only hold the amulet once, and if no players remain that may hold the amulet, it shatters, killing the current holder. The Amulet Holder may shatter the amulet destroying it and killing the holder.

PR #691, opened 2022-07-06
Correction regarding UAs

ThomasRubycrust:
Replace with:
> An exception is Unaligned players, the werewolves do not need to kill them, however any Unaligned player that remains alive at the end of the game, but has not hit their win condition will still lose.


venomousbirds (approved):
Solo doesn't have this description?

McTsts:
Forgot Solo existed tbh, but just checked and it doesn't seem to be anywhere else, but I did find some other issues (#693)

PR #729, opened 2022-09-22
Update bard

ThomasRubycrust:


McTsts:
That could also work. It still means the Bard itself confirms itself, but I don't know how we could stop this...

Issue #732, opened 2022-09-22
Role Suggestion: Alcoholic


McTsts:
The Alcoholic spends every night at the tavern so they are immune to all attacks (as they arent home), but can still die by other means. At the end of the first night without a living Bartender the Alcoholic dies.

A role to allow a balanced Bartender. With an Alcoholic (or the possibility of one) ingame the Bartender cant immediately reveal and gather claims or risk losing a powerful town role.

cryptidjayden:
makes sense to me

McTsts:
PR was created for this: #737

McTsts:
This is the role description so far:

**Alcoholic** | Townsfolk Miscellaneous
__Basics__
The Alcoholic spends the night drinking at the tavern and is thus absent for any attacks on them.
__Details__
During the night, the Acoholic is fully immune against all attacks targeting them, though they may still be killed by other means.
The Alcoholic cannot exist without a bartender, and loses their power as soon as the bartender dies. The Alcoholic dies at the end of the first night without a living Bartender.

__Simplified__
At night, the Alcoholic is immune against all attacks, but dies when the Bartender dies.

PR #737, opened 2022-09-23
Role Suggestion: Alcoholic

McTsts:
Implements #732


cryptidjayden (approved):

ThomasRubycrust:
Role description at top.

❌ Reject
✅ Approve

ThomasRubycrust:
So the point of this role is to make Bartender usable. Bartender can give town advice and confirm themselves to trusted town members but if they reveal themself fully they will also take down the Alcoholic. My alternative suggestion would've been to remove Bartender as I just think it's power is too unbalanced otherwise (and I think this shows - its IIRC the town role that has been unused for the longest amount of time)

PR #753, opened 2022-10-01
Nerf Stalker

### LINKED ISSUE #599: Buff Stalker

McTsts:
Fixes #599

### DIFF
--- a/townsfolk/power/stalker
+++ b/townsfolk/power/stalker
@@ -1,8 +1,8 @@
 **Stalker** | Townsfolk Power
 __Basics__
-Each day, the Stalker may assign two extra votes to a player in the current day’s polls.
+Each day, the Stalker may assign two extra invisible votes to a player in the current day’s polls.
 __Details__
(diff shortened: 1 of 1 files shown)


McTsts:
That's why I want to buff it to make it a bit more useful :)

McTsts:
>  **Ethan.**: What if we made stalker something else

I'm not sure if we should introduce a new Stalker role to replace the old Stalker, it's one of the less fitting names IMO. Also we had a PR to remove Stalker a while ago, but then decided to keep it.

McTsts:
I think it was Vera

McTsts:
https://github.com/McTsts/Werewolves-Roles/pull/621

McTsts:
Well then we said the role needs a buff - here's a buff.

McTsts:
That makes more sense as a new role then in my opinion. Adding a new name with a weak name seems a bit pointless.

PR #755, opened 2022-10-01
Create Flute Charmer

12Metatron:


McTsts:
For bigger games: If there are both a Flute Charmer and Flute Apprentice and the Flute Player dies, do both promote into a Flute Player, or does only one do so (and if this is the case, which one does).

12Metatron:
> For bigger games: If there are both a Flute Charmer and Flute Apprentice and the Flute Player dies, do both promote into a Flute Player, or does only one do so (and if this is the case, which one does).

I'm not sure. It'd be whatever the edge case is with having 2 flute apprentice

McTsts:
Apparently, there isn't a solution for that edge case. I guess right now the answer would be that both promote.

McTsts (approved):

McTsts:
Fixed it

McTsts (commented):
Testing if reviews now get sent to discord.

McTsts:
**Flute Charmer** | Solo Power - Flute Team | Limited
__Basics__
The Flute Charmer is a flute player that can’t enchant other players; instead, they have the ability to charm one player each day, learning their role and disguising them as flute player. The Flute Charmer becomes a flute player if none remain.
__Details__
During the day, the Flute Charmer may charm a player, the charmed player is not informed about this. This action is immediate. 
Successfully charming a player will tell the Flute Charmer their targets role. This result is affected by strong disguises, redirections, and obstructions. Charmed players are weakly disguised as flute player.
This effect will disappear once no members of the Flute Player channel or team remain. The players are not told when the effect disappears.
When a player is both charmed and enchanted, the charmed status is removed.
The Flute Charmer can target themself or other members of the flute team.
If no flute players remain, the Flute Charmer turns into a flute player and loses their current powers.

__Simplified__
The Flute Charmer becomes a flute player if none remain.
Additionally, once per day, the Flute Charmer can learn the role of a player and disguise them as flute player.
An enchanted player will lose the disguise given to them by Flute Charmer.

❌ Reject
✅ Approve

PR #770, opened 2022-10-04
Delete scared white werewolf

McTsts:


12Metatron:
I feel like it's used almost as much as regular www, in recent memory. But I could be mistaken

PR #777, opened 2022-10-07
Fix FPxCupid edge case

### LINKED ISSUE #773: Plague Team X Cupid Bug
**Ethan.**:
I just noticed that currently if a Plague team member and a Couple consisting of the other plague team member and a plagued cupid
The plague team can never win on their own abilities and must essentially throw, because they can't be sure the other plague member is cupided, and get the other plague member lynched

Idk if this is intentional or not

### LINKED ISSUE #776: Flute Team x Cupid
**Ts**:
As mentioned in #773 There's an issue when several flute team members, one with a cupid remain. The non lover flute team member then has no way of winning as they'd need to enchant the lover which is impossible. As explained by Ethan in #773 that's different from how other couples work and should be fixed.

McTsts:
Fixes #776

### DIFF
--- a/other/flute/flute player
+++ b/other/flute/flute player
@@ -1,7 +1,7 @@
 **Flute Player** | Solo Power - Flute Team
 __Basics__
 Each night, the Flute Player may enchant a certain amount of players. 
-If all living players are either enchanted, unaligned or Flute team members, the Flute team wins.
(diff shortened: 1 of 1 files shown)


McTsts:
I mean #776 not #773

McTsts:
I don't think there's really anything to discuss here, so I'll merge it.

PR #780, opened 2022-10-08
Make Reaper SK

### LINKED ISSUE #757: Should Wraith stay as SK?
Reaper is considered SP and Wraith is SK even though its killing ability is one that Reaper shares. 

@12Metatron brought up that for solos it makes sense to prioritize abilites that correspond to their win condition. 
 So in this case: Reaper & Wraith _can_ kill, but it's not their goal. They can also both use a _power_ and using their power on everyone is their goal so it makes more sense to have that as the category.

McTsts:
Fixes #757


McTsts:
Changed this to make it Reaper SK instead

venomousbirds (commented):
Makes sense with what it's being compared to (Devil being Killing too even though its main goal isn't to kill). I guess what might make it sound even more like a killing role is to add that one part that a lot of other solos have with its win condition: "...wins when all living players have an effect applied, are part of that solo team or are dead"
So maybe adding the "or are dead" part would make it more obviously solo killing?

McTsts:
I can't decided whether I support SP or SK

cryptidjayden:
i agree with shapechange

McTsts:
Me too

ThomasRubycrust:
I'm with Ts

PR #825, opened 2022-11-03
Buff Bat

ThomasRubycrust:
Took out the part limiting who they can target. Role isn't strong enough to need that limitation.


McTsts:
This needs to be recreated

PR #908, opened 2023-01-09
Create Coyotes

Shepherdwhen:


rupertandrosy:
changed coyote pup to lesser coyote bc pup sounds like a cub/child clone

limited/Coyote Pup
```diff
@@ -1,4 +1,4 @@
-**Coyote Pup** | Werewolf Group
+**Lesser Coyote** | Werewolf Group
 __Basics__
 The Coyote Pup is a packless non-lycan wolf who shares a secret channel with the other coyotes: however, the coyotes do not know the identities of any other wolves.
```
12Metatron, line 3: changed role name but desc still mentions old name

rupertandrosy:
this needs the roles added to the lists, and then i don't see any reason it couldn't be merged

PR #979, opened 2023-02-17
Update rogue wolf

Shepherdwhen:
Makes rogue wolf less punishing for having no info while keeping it more unique and interesting


rupertandrosy:
i would not call this role particularly "interesting" but at least it's a little more balanced now, i suppose.

PR #1076, opened 2023-06-10
Make innkeeper consistent

Shepherdwhen:
remove the strong disguises to make inkeeper consistent with other roles


McTsts (changes requested):
I feel like consistency is detrimental to balance here

PR #1156, opened 2023-07-21
Bakers Nerf

McTsts:


rupertandrosy (approved):
one comment - i see we've added an edge case section to this role. are we committing to doing that for the other roles too? @McTsts

McTsts:
@rupertandrosy it turns out we were already doing it - at least for Infected Wolf

PR #1162, opened 2023-07-24
Update Dog & Cat

Shepherdwhen:


McTsts:
Cant figure out how to get rid of the extra commits, so merging like this

rupertandrosy:
summary of changes:
- spag fixes (obviously)
- dog/cat that becomes warlock/ft can't use N1 check. this is so that if dog/cat becomes rival, they can more easily fake warlock/ft

@McTsts

rupertandrosy (approved):

PR #1177, opened 2023-08-12
Delete hellhound

Shepherdwhen:


McTsts (changes requested):
Needs removal from various lists

McTsts (approved):

PR #1187, opened 2023-10-09
New em roles

12Metatron:


extra/miscellaneous/offer recipient
rupertandrosy, line 1: If we don't delete Cerberus, this file will need editing, but I won't do it yet, in case Cerberus is deleted.

rupertandrosy (commented):
Updated grammar for the new EM roles*. Change the name from Journal -> Journal Recipient, because that was stupid. Update mentions of relevant EM roles in Devil, Devil-relevant, and Chef descriptions. 

*Have NOT updated the OR description. Have NOT made any changes to the Journalist description. This PR should not be merged until we choose whether to delete Cerberus or not.

12Metatron:
to add(?): raven keeper and baroness

PR #1224, opened 2023-11-04
Make PD nonaligned and remove 'contagious' as an effect

McTsts:


rupertandrosy (approved):
This PR now also removes the effect 'contagious'. All the consequences of the contagious effect are now consequences of the 'contaminated' effect.

PR #1228, opened 2023-11-17
Formalize Everything

McTsts:


McTsts (commented):
I dont think role change applies to groups. New type of role change called group change?

PR #1234, opened 2023-11-18
CL executes Cult kill

McTsts:


McTsts:
need changes: update the ability def

PR #1260, opened 2023-11-30
Formalize ability types

McTsts:


McTsts (commented):
reading all these after each other I realize redirection and killing defs conflict: Killing defines a killing is a success if anyone dies, but redirect says its a failure if it gets redirected (which still may lead to a death)

12Metatron:
oh yeah, should mention transformation-type roles in the role definition

PR #1274, opened 2023-12-03
Delete Recruitment Category

venomousbirds:
I probably did this completely wrong but I tried


McTsts (approved):
changed everything i could find

PR #1296, opened 2023-12-13
Remove new TI from baroness roles

Shepherdwhen:


McTsts (changes requested):
Tanner is once again WP

rupertandrosy:
i maintain that it would be easier if baroness had no special roles, but i have nothing specific to review here

Shepherdwhen:
> Tanner is once again WP

@McTsts Done

McTsts (approved):

PR #1310, opened 2023-12-18
Re Add Cerberus

Shepherdwhen:


McTsts (changes requested):
Needs a card description

McTsts:
Idea was cancelled as discussed in #1301

PR #1323, opened 2024-01-17
Create unaligned master

Shepherdwhen:
A list of all unaligned roles rather than the split categories just for ease of access


McTsts (approved):

McTsts:
marked limited roles with * as we usually do. Didnt check list contents but im gonna trust on you on them

PR #1351, opened 2024-04-27
Monarch Changes

McTsts:
- Merge Knight into Heir
- Make Heir selection a forced ability


venomousbirds (changes requested):
Not a big fan of the continuity of Monarch, I think adding a way to lose the role entirely might be more balanced in the sense that it's not a constant threat all game long. Maybe by removing the random player thing?

venomousbirds (approved):

PR #1356, opened 2024-04-30
Reaper / Wraith are not sleepless

McTsts:
Reaper was specified to be sleepless, but with what we have we'd need to ensure its permanent, as otherwise an Exorcist could revoke the effect, excluding the reaper from their own win condition

As this is very complicated, I have instead opted to remove the sleepless effect from reaper/wraith and added the nightmare alignment back into the team win condition


venomousbirds (approved):

McTsts:
Merging this anyway due to potential thief's hell usage

PR #1358, opened 2024-05-01
Delete Nightmare Team

McTsts:
Nightmare team has repeatedly failed to satisfy us, so its time to delete it


venomousbirds (approved):
I just approved a change to this team but also it's a terrible team

PR #1367, opened 2024-06-15
Haunting Game Mode

McTsts:
Implements #1007 

Was originally discussed a while ago, new discussion is in #planning


ThomasRubycrust (approved):
Approving for v18

PR #1378, opened 2024-08-06
Bitter Hag Formalization Fix

McTsts:
Ability set `Coven` should be specified as a string


McTsts:
Bitter Hag still wont parse even with this change as apparently Coven itself can't be parsed, but it should now have the same error as other hags which fail due to invalid ability set and not due to incorrect syntax

PR #1385, opened 2024-09-13
Update flute apprentice

12Metatron:
charmed players are affected


McTsts (approved):
Seems good to me if it gets approved

PR #1386, opened 2024-09-13
unlimit and change promotion order

12Metatron:
unlimit flute charmer and change promotion order

### DIFF
--- a/other/flute/flute charmer
+++ b/other/flute/flute charmer
@@ -4,7 +4,7 @@ The Flute Charmer is a flute player that can’t enchant other players; instead,
 __Details__
 During the day, the Flute Charmer may charm a player, the charmed player is not informed about this. This action is immediate. 
 Successfully charming a player will tell the Flute Charmer their targets role, and weakly disguise them as flute player afterwards. This result is affected by all disguises and redirections. 
-This effect (including the disguise) will disappear once no members of the Flute Player channel or team remain. The players are not told when the effect disappears.
+This effect (including the disguise) will disappear once no members of the flute player channel or team remain. The players are not told when the effect disappears.
 When a player is both charmed and enchanted, the charmed status and disguise is removed.
 The Flute Charmer can target themself or other members of the flute team.
 If no other flute team roles remain, the Flute Charmer turns into a flute player and loses their current powers.
--- a/other/flute/flute charmer
+++ b/other/flute/flute charmer
@@ -19,7 +19,7 @@ Starting: Join #Flutes
 Immediate Day: [Attribute: @Selection lacks `Enchanted`] |flute_charmer.1|
   • Role Investigate @Selection (SD, WD)
   • Apply `Charmed` to @Selection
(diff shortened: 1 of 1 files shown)


MrTurtleJr:
Line 13 currently reads "The Flute Charmer becomes a flute player if The Flute Charmer becomes a flute player if no other flute team roles remain." as far as I can see - that should be changed.

MrTurtleJr:
Considering that both the Flute Player and the Flute Apprentice capitalize every instance of 'Flute Player' (instead of flute player) and 'Flute team' (instead of flute team), perhaps this could be done with the Flute Charmer as well in the name of standardized capitalization/orthography?

McTsts:
> Considering that both the Flute Player and the Flute Apprentice capitalize every instance of 'Flute Player' (instead of flute player) and 'Flute team' (instead of flute team), perhaps this could be done with the Flute Charmer as well in the name of standardized capitalization/orthography?

we usually aim to only capitalize the current role / game element. I dont think capitalizing solo primary in solo secondary is something we do for other teams (or do we? should we?)

MrTurtleJr:
> > Considering that both the Flute Player and the Flute Apprentice capitalize every instance of 'Flute Player' (instead of flute player) and 'Flute team' (instead of flute team), perhaps this could be done with the Flute Charmer as well in the name of standardized capitalization/orthography?
> 
> we usually aim to only capitalize the current role / game element. I dont think capitalizing solo primary in solo secondary is something we do for other teams (or do we? should we?)

Oh okay. That's fair, I saw that with Hell team, where Demon does not capitalize devil. However, I was going off of FPA description, which does do this. In any case, then that would have to change. 
However, what about Flute team? From what I can tell, it is capitalized in the FP and FPA descriptions. For Devil and Demon it seems to be roughly 50/50. Are alignments capitalized?

McTsts:
> > > Considering that both the Flute Player and the Flute Apprentice capitalize every instance of 'Flute Player' (instead of flute player) and 'Flute team' (instead of flute team), perhaps this could be done with the Flute Charmer as well in the name of standardized capitalization/orthography?
> > 
> > 
> > we usually aim to only capitalize the current role / game element. I dont think capitalizing solo primary in solo secondary is something we do for other teams (or do we? should we?)
> 
> Oh okay. That's fair, I saw that with Hell team, where Demon does not capitalize devil. However, I was going off of FPA description, which does do this. In any case, then that would have to change. However, what about Flute team? From what I can tell, it is capitalized in the FP and FPA descriptions. For Devil and Demon it seems to be roughly 50/50. Are alignments capitalized?

The only specific captialization rule we formally recognize is doing it for the current role's name. We could do it for solo alignments (from their solo roles) and solo primaries (from their secondaries) as well which would make sense, but this is currently not defined. (Should we change this?)

Right now we just have:

"In a role's description the name of the role itself is capitalized, but other role names are not." defined over [here](https://github.com/WerewolvesRevamped/Werewolves-Roles?tab=readme-ov-file#public-information)

McTsts (approved):

PR #1395, opened 2024-09-16
buff apothecary

12Metatron:
implements #1370

### DIFF
--- a/limited/apothecary
+++ b/limited/apothecary
@@ -3,9 +3,10 @@ __Basics__
 Each night, the Apothecary may poison a player, killing them at the end of the next night. Alternatively, they may cure the player they last poisoned.
 __Details__
 The Apothecary starts the game with three flasks of poison and a single antidote.
-When the Apothecary poisons a player, that player will be killed at the end of the next night (this is not an attack). The Apothecary will only poison a player if they survive the night. 
+When the Apothecary poisons a player, that player will be killed at the end of the next night (this is not an attack). The Apothecary will only poison a player if the Apothecary survives the night. 
 Instead of poisoning a player, the Apothecary may also use their antidote to cure the player they have previously poisoned, saving them from death.
 Players are not notified of being poisoned or cured.
+If a targeted player dies the same night as being poisoned or cured, the flasks are not consumed.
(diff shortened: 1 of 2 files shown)


McTsts (changes requested):
Unclear who "they" refers to in "if they survive the night". The Apothecary or the poisoned player?

Unsure how to rephrase it

McTsts (changes requested):
I do not think the new change with flasks not being consumed works

McTsts (approved):

McTsts (approved):
what

McTsts (changes requested):
what 2

limited/apothecary
```diff
@@ -28,6 +28,7 @@ Choice `Cure` Chosen:
   • Remove `CureAvailable` from @Self
 Passive Start Day:
   • Change `Poisoned` value `1` to `Active` for @(AttrSelf:Poisoned)
+On Death: Remove `Poisoned:Self` from @(AttrSelf:Poisoned)
```
McTsts, line 31: why did we add this, this would make it so anyone poisoned would lose it when Apothecary dies when we are trying to implement the opposite

limited/apothecary
```diff
@@ -3,7 +3,7 @@ __Basics__
 Each night, the Apothecary may poison a player, killing them at the end of the next night. Alternatively, they may cure the player they last poisoned.
 __Details__
 The Apothecary starts the game with three flasks of poison and a single antidote.
-When the Apothecary poisons a player, that player will be killed at the end of the next night (this is not an attack). This will only happen if the Apothecary survives until the end of that night. 
+When the Apothecary poisons a player, that player will be killed at the end of the next night (this is not an attack). The Apothecary will only poison a player if the Apothecary survives the night.
```
McTsts, line 6: Then why is it even an immediate ability? No this doesnt make any sense. If anything the poison should kill no matter what. It's immediately applied and then death is sure (unless cured or protected). This would be a further buff but I dont see the problem as this doesnt make it stronger than roles as Assassin still

McTsts (changes requested):
I'm confused with what we are doing in this PR

PR #1460, opened 2024-12-01
Exorcist x FP Clarification

McTsts:
Clarify that channel joining is not undone by Exorcist


McTsts:
This is already suggested to work this way here imo:

> Whenever a player is enchanted, they will immediately join a secret channel with the other enchanted players, but without the Flute Player(s).

So I would not consider this a change

McTsts:
Fixes #1400 partially

rupertandrosy (approved):

PR #1473, opened 2024-12-18
stalker better phrasing

12Metatron:


McTsts (approved):

Shepherdwhen (changes requested):
I don't think it needs to be starting Day 2, keep it as After Day 1 but just add the "Each day," part before it, this keeps it the same as every other role with a day limited clause

I also don't think you need the if there are more than one lynch polls part just change it to "That player will have two more invisible votes against them on the lynch polls that day"

PR #1539, opened 2025-04-21
Remove vote negation from Monarch

### LINKED ISSUE #1531: Monarch
**mctsts**:
Does monarch feel too strong?

McTsts:
Fixes #1531

### DIFF
--- a/_game/attributes/heir
+++ b/_game/attributes/heir
@@ -3,7 +3,5 @@ __Basics__
 Applied by the Monarch. Grants an additional vote on polls, and negates a vote.
 
 __Formalized__
-Passive Start Day:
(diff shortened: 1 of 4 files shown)


dahliazhou13 (approved):
LGTM

chimilovesshrimp (approved):
LTGM

Shepherdwhen (commented):
Why does it need to keep reapplying every day, we could just have special public power be active as long as the player has the attribute

rupertandrosy (approved):
Text-wise, this is fine. No more text reviews are needed. @Shepherdwhen @McTsts if/when the formalisation is ready, this can be merged.

McTsts:
> Why does it need to keep reapplying every day, we could just have special public power be active as long as the player has the attribute

I really cannot think of any reason..... I think we can safely change to just Starting trigger only

PR #1544, opened 2025-04-24
Change Cursebearer

Shepherdwhen:


rupertandrosy (changes requested):
It's not clear from the description whether CB can be linked to a WWW. I would suggest that seems bad. Whether it can or cannot be, this should be explicit in the description (probably either through "an arbitrary member of the wolfpack of any alignment", or "an arbitrary wolf-aligned member of the wolfpack".

McTsts:
> It's not clear from the description whether CB can be linked to a WWW. I would suggest that seems bad. Whether it can or cannot be, this should be explicit in the description (probably either through "an arbitrary member of the wolfpack of any alignment", or "an arbitrary wolf-aligned member of the wolfpack".

Werewolves are by definitions members of the werewolf team, so I guess technically WWW is not a werewolf, though it wouldn't hurt to clarify this

_game/attributes/cursebearerlink
```diff
@@ -0,0 +1,9 @@
+**Cursebearer Link** | Attribute
+__Basics__
+An attribute marking a wolf that is linked to a cursebearer.
+
+__Formalized__
+Starting: 
+  • Target @ThisAttr->Source
+  • Reveal `@Self is @Self->Role and your linked wolf` to @Target
```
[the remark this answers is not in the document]
McTsts: I have removed this and skipped the targeting since that seems like it doesn't really make sense

limited/cursebearer
```diff
@@ -1,18 +1,20 @@
...
-If no players with that role remain, the Cursebearer can't use its powers anymore, but retains the disguise.
+__Basics__
+The Cursebearer is a member of the townsfolk team who is disguised as an arbitrary werewolf. If the werewolf they are disguised as dies, the Cursebearer is informed.
 
 __Formalized__
 Role Attribute: `Lycan`
-Starting: Strongly Disguise @Self as %Role%
+Starting: 
+  • Target %Role% (Role)
+  • Apply `CursebearerLink` to @Target->RandomPlayer
```
McTsts, line 16: There is absolutely no way this works, right? Is there?

McTsts (approved):
Has this been tested?

rupertandrosy (commented):
approved on my end

limited/cursebearer
McTsts: According to [WWRF](https://github.com/WerewolvesRevamped/Werewolves-Roles/blob/main/WWRF%20Guide.md#role-type) roles support a "Players" property access, but not a "RandomPlayer" one

McTsts (commented):
Requires testing

rupertandrosy (approved):

McTsts (changes requested):
See #1540

rupertandrosy:
> See #1540

resolved now? @McTsts

McTsts:
yup, resolved by #1540

PR #1551, opened 2025-04-26
Change Coyotes

### LINKED ISSUE #1481: Coyote change
**captainluffy**:
Make it so Coyotes don't lose their power when group is disbanded. With the current roles this has no functional change, but I think it's a good idea in general

McTsts:
Coyotes no longer lose their power
Fixes #1481

### DIFF
--- a/_game/groups/coyotes
+++ b/_game/groups/coyotes
@@ -1,6 +1,6 @@
 **Coyotes** | Werewolf Team Group
 __Basics__
-The Coyotes are a secondary wolf group, containing (usually) one Greater Coyote. The coyotes and the wolfpack do not know each other, but have the same alignment. If the wolfpack fails to kill a townsfolk, the coyotes can attack a player in the next night, but if the Greater Coyotes dies, they lose their ability and join the wolfpack instead.
+The Coyotes are a secondary wolf group, containing (usually) one Greater Coyote. The coyotes and the wolfpack do not know each other, but have the same alignment. If the wolfpack fails to kill a townsfolk, the coyotes can attack a player in the next night, but if the Greater Coyotes dies, they coyotes are disbanded and they join the wolfpack instead.
 __Members__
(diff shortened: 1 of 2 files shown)


_game/groups/coyotes
```diff
@@ -1,6 +1,6 @@
 **Coyotes** | Werewolf Team Group
 __Basics__
-The Coyotes are a secondary wolf group, containing (usually) one Greater Coyote. The coyotes and the wolfpack do not know each other, but have the same alignment. If the wolfpack fails to kill a townsfolk, the coyotes can attack a player in the next night, but if the Greater Coyotes dies, they lose their ability and join the wolfpack instead.
+The Coyotes are a secondary wolf group, containing (usually) one Greater Coyote. The coyotes and the wolfpack do not know each other, but have the same alignment. If the wolfpack fails to kill a townsfolk, the coyotes can attack a player in the next night, but if the Greater Coyotes dies, they coyotes are disbanded and they join the wolfpack instead.
```
chimilovesshrimp, line 3: Should we clarify that they don't lose their ability?

Shepherdwhen: By default if role change isn't mentioned or ability loss isn't mentioned it's assumed they keep the ability

Shepherdwhen (approved):

rupertandrosy:
in terms of text i approve this. however, i think there may be an issue which i will drop in the discord thread

_game/groups/coyotes
McTsts: I would say we would clarify if this was a change that people wouldn't be aware off, but since there have never been coyotes with powers its probably fine? Though we do have foxes and lone werewolves, both of which do lose powers on pack join, so maybe a clarification doesn't hurt.

McTsts:
Looks like this is good to merge.

PR #1553, opened 2025-04-27
New Role: Farmer

McTsts:
Implements #1456


dahliazhou13 (commented):
Some general suggestion for farmer can be applied to all three roles. Just some nitpick

rupertandrosy:
i have made changes to certain parts of the "Farmer" wording. these need to be applied to fw and scarecrow too - if nobody else wants to do this, i can do it myself at a later date. also, there were role capitalisation issues in farmer that i expect are also present in fw and scarecrow.

rupertandrosy (approved):
augh

PR #1554, opened 2025-04-29
Fix devil

### LINKED ISSUE #1507: Make wager kills fail on soulless targets instead of asking you to change targets.

McTsts:
Fixes #1507


rupertandrosy:
i will approve this bc i assume youve got it right. ig if u want an actual review get ethan to do it

rupertandrosy (approved):
approved, needs testing

PR #1555, opened 2025-04-29
Create horrifying wolf

RatonGolzache:


rupertandrosy (dismissed):
the horrifying wolf is veri spooki

limited/horrifying wolf
```diff
@@ -0,0 +1,15 @@
...
+This applies even if a disguise or obstruction affects the investigation.
+
+__Simplified__
+Whoever perceives the Horrifying Wolf may not investigate the following night. 
+
+__Formalized__
+Inherit: `Pack Lycan`
+
+__Card__
+The Horrifying Wolf is veri spooki.
```
McTsts, line 15: no

McTsts (changes requested):

McTsts:
I have added formalization for the role

HW:
On Visited [Investigating]: Process: Apply `Horrified` to @Visitor (~NextNight)

Horrified:
Passive Start Day: Obstruct Investigating for @Self (~NextNight)



It grants horrified immediately on the visit, behind a Process, so it doesn't learn anything.
We use horrified as the effect should only start working after the end of the current night (Passive Start Day instead of Passive End Night to avoid conflicts with end night investigations which I don't think exist currently, but could in theory)
We cannot have the end of phase application logic in HW as HW could be killed immediate/end night and the effect still occurs.

Is obstructing investigating like I did sufficient to achieve what we want? How does this work with roles like Aura Teller or Crowd Seeker? What do these do with this?

McTsts (commented):
Text good, Formalization needs testing

McTsts (dismissed):
Tested it and it works for FT as expected, but we may need to adjust roles like AT/CS for them to make sense with this

McTsts:
Adjusted some investigation, needs testing

McTsts:
I have just tested with CS and seems to work as intended
image

McTsts:
The other roles match CS so they should be good as well

McTsts (approved):
LGTM

RatonGolzache:
Can this give an ominous message when the investigation failed or is that not supported

McTsts:
> Can this give an ominous message when the investigation failed or is that not supported

Would probably be a bit awkward to do, though this also means it would be easily differentiable from other role blocking

PR #1559, opened 2025-04-30
Aligned

McTsts:
Add "-" to "<team>-aligned" where it is missing


McTsts:
These are the only instances I could find

rupertandrosy (approved):

PR #1563, opened 2025-04-30
Two new wolfpack roles

rupertandrosy:
Previously known as "ritualistic" and "pursuing" wolf. Names open to suggestions. This is for me to work off of.

### DIFF
--- a/werewolves/killing/ferocious wolf
+++ b/werewolves/killing/ferocious wolf
@@ -1,0 +1,14 @@
+**Ferocious Wolf** | Werewolf Power
+__Basics__
+
+__Details__
+
+
+__Simplified__
+
+
+__Formalised__
+
+
+__Card__
+The Ferocious Wolf hates to see a victim escape its grasp.
--- a/werewolves/power/mother wolf
+++ b/werewolves/power/mother wolf
@@ -1,0 +1,14 @@
+**Mother Wolf** | Werewolf Power
+__Basics__
+
+__Details__
+
+
+__Simplified__
+
+
+__Formalised__
+
+
+__Card__
+The Mother Wolf is very protective of her pack.


McTsts:
- [ ] Added to wolfpack list https://github.com/WerewolvesRevamped/Werewolves-Roles/blob/main/_game/groups/wolfpack

rupertandrosy:
role text done. @McTsts @kruthers over to you for formalising

i may accept alternative role name suggestions but i am fairly happy with the existing ones

a v20.1 project

McTsts:
Ugh, this is gonna be annoying. I think we need to built something into the wolfpack poll to make FW work

rupertandrosy:
> Ugh, this is gonna be annoying. I think we need to built something into the wolfpack poll to make FW work

is the issue checking if the attack fails or making sure the attack wasn't on the fw selection?

McTsts:
Its a similiar case to the coyote logic which has to be built into the wolfpack poll, so we have to add logic into the WP poll to check if the attack was successful (we can check the feedback of the attack action so no problem), if that's the case we can do something like "Emit `Attack` for @(Role:Ferocious-Wolf)" which is basically calling a function on the FW, the FW has a trigger that listens for that via "On `Attack` Emitted: ..." and does the rest of the logic there

rupertandrosy:
i no longer like the name mother wolf bc i think it reinforces the patriarchy. will think of a new name

rupertandrosy:
ritualist wolf is better than the other ritual-based names but still seems....... annoying
seer wolf is okay, or maybe diviner wolf
gambler wolf (lets go gambling) (not a serious suggestion)

McTsts:
Formalization for FW is UNTESTED

McTsts:
Ferocious Wolf seems good I think

McTsts:
ritualist wolf seems good too, but relies on this fix https://github.com/WerewolvesRevamped/Werewolves-Bot/commit/bf3e233314bbbbb0988469a6d481b243f32fc776

McTsts (approved):
LGTM

PR #1570, opened 2025-05-13
Create raccoon

12Metatron:


limited/raccoon
```diff
@@ -0,0 +1,17 @@
+**Raccoon | Unaligned Power | Limited**
+__Basics__
+Each day the Tanuki may strongly disguise two players into a role of their choice until the next day phase.
```
chimilovesshrimp, line 3: a role, does it mean both players will be disguised to the same role?

chimilovesshrimp: Also should we align tanuki vs racoon?

12Metatron: Good catch, I'll go fix it

chimilovesshrimp (approved):

McTsts:
I have put a formalization, but it needs testing

PR #1572, opened 2025-05-18
Vanishing prompts

McTsts:
Adds vanishing prompts and uses them for the 3 base elected roles


McTsts:
Should the prompts for roles like Assistant also be set to vanishing?

PR #1581, opened 2025-06-09
Coyote Fix

McTsts:
This should make it so coyotes are actually disbanded on the GC death leading to the other coyotes joining WP on disbandment


McTsts:
Pushing without approval as no text changes are in this PR

PR #1618, opened 2025-10-05
Allow untargetting

McTsts:
This implements a bot change from https://github.com/WerewolvesRevamped/Werewolves-Bot/issues/73 to allow untargetting


McTsts:
@WerewolvesRevamped/rule-book-committee Can anyone think of a role that should be able to untarget besides Huntress, Assistant, Ranger, Angel and Farmers? If no please approve

PR #1628, opened 2025-10-12
Big bug change

### LINKED ISSUE #1485: Firebug is OP
**mctsts**:
Firebug is imo too strong. It isn't too strong directly though - its warding can be avoided in many cases. The issue doesn't lie in its ability to kill/powder/ignite, its issue lies in its ability to deny many other strategies. This is both too powerful and reduces the overall potential of strategies and pushes the game more towards randomization, which is bad.

Let's look at some scenarios:
• There is no clear kill target. The Firebug will likely ward somewhat unpredictably and wolves will kill randomly/unpredictably so there is no issue here
• There is a clear kill target. Firebug will likely ward the clear kill target and wolves will assume this is what was likely done and will pick another target. Best case this is another solid kill candidate (but then this would fall under the next scenario), worst case the kill target is more random which is not really a desirable outcome. The Firebug may also predict that the wolves will assume the clear kill target is warded and thus ward somebody else, but this is either the next scenario or random/unpredictable. Wolves can't really afford to risk attacking the clear kill target even if they think Firebug may use this logic as the 1p loss to wolf team generally really hurts
• There are two clear kill targets. The wolves can expect Firebug to ward one of them and can't really afford to risk attacking either of them as once again the risk of hitting the ward is too high. They must thus select another target
• There are more likely kill targets. If there's 3 or more than wolves can maybe risk it

This assumes both Firebug and Werewolves analzye which people would make good kills and Werewolves consider that Firebug is in game and thinking strategically.

Overall the effect of Firebug is thus to decrease strategic kills and increase more random kills.

I think in this regard the role is too strong and negatively impacts the game.

rupertandrosy:
Fixes #1485


McTsts (approved):
Tested, seems to work

PR #1629, opened 2025-10-12
Fix monarch

rupertandrosy:
NOTE THE BASE BRANCH

rewrite is done. two things:
1. emergency change may not be formalised (see #1626)
2. im not sure if the forced heir is completely forced rn: it says that it's passive/persistent, but the first forced selection only happens at phase end, bringing up the question of whether it would pick and then succeed to an heir if the monarch died mid-n1


McTsts:
I added a fix for 2. and tested both changes

McTsts (changes requested):
Would you be able to update the phrasing some more? I think it would be important to contain that this is an elected role that by default is NOT part of the game and only exists when specifically added to the game. Not sure how to best phrase this.

PR #1670, opened 2025-12-02
GF Fix

### LINKED ISSUE #1637: Gallant Fox
@rupertandrosy I undid your change to GF as it as technically difficult to execute, so since I assume you want it done anyway, here's an issue for it.

McTsts:
Fixes #1637


McTsts:
@WerewolvesRevamped/rule-book-committee If GF chooses a target but the target dies mid-day nothing is announced to wolfpack. Everyone fine with that?

